PR #1, opened 2022-01-31
Read RSSI from a connected BLE device

sophisticode:
This change adds functionality to read RSSI from a connected device. All required logic to iOS and Android native code is implemented and protos have been updated and regenerated for the required data structures
The example app was also updated to periodically show the RSSI of a connected device.
Furthermore, the regenerate.md file from flutter_blue was copied and flutter_blue_plus related changes were done to help anyone when doing future changes to protos.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ For location permissions on iOS see more at: [https://developer.apple.com/docume
 | state                       |  :white_check_mark:  |  :white_check_mark:  | Stream of state changes for the Bluetooth Device. |
 | mtu                         |  :white_check_mark:  |  :white_check_mark:  | Stream of mtu size changes. |
 | requestMtu                  |  :white_check_mark:  |                      | Request to change the MTU for the device. |
+| readRssi                    |  :white_check_mark:  |  :white_check_mark:  | Read RSSI from a connected device. |
 
 ### BluetoothCharacteristic API
 |                             |       Android        |         iOS          |             Description            |
--- a/android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -659,6 +659,24 @@ public class FlutterBluePlusPlugin implements FlutterPlugin, MethodCallHandler, 
         break;
       }
 
+      case "readRssi":
+      {
+        String remoteId = (String)call.arguments;
+        BluetoothGatt gatt;
+        try {
+          gatt = locateGatt(remoteId);
+          if(gatt.readRemoteRssi()) {
+            result.success(null);
+          } else {
+            result.error("readRssi", "gatt.readRemoteRssi returned false", null);
+          }
+        } catch(Exception e) {
+          result.error("readRssi", e.getMessage(), e);
+        }
+
+        break;
+      }
+
       default:
       {
         result.notImplemented();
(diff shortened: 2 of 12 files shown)


maciej-rexton (commented):
Hi, I really like this pull request, as I've missed this functionality lately. 

But I'm curious, if it is possible to implement this in "reactive" style. I mean that instead of calling `readRssi` method, there would be `Stream<int> rssi` field, which would emit a new event every time the rssi changes.

I see that the last changed lines from `example/lib/main.dart` try to imitate that, by just getting the value periodically. But I mean something more like, being notified about rssi change by the android SDK instead of polling.

I'm asking, because I don't have much experience in that topic, but the `rssi` field would match better the rest of the library (like `state` fields in `FlutterBluePlus`/`BluetoothDevice` or `isScanning` etc.) and I think it would be better to implement it like that if it's possible.

sophisticode:
The functionality was added as closely as possible to how it works on the native side. This gives the app the highest flexibility how it wants to use the functionality (e.g. if it needs the value once to be able to warn in case of weak signal before doing some BLE communication, or if it needs it regularly e.g. to display the signal strength in the UI). 

Providing a stream of rssi values by the plugin would mean, that several assumptions have to be made. E.g. the frequency of reading the rssi, post-processing (e.g. low-pass filtering as the value fluctuates strongly for single measurements), decision when a new rssi is different enough (e.g. the app might want to know only if the value is within a certain range) are different depending on the use-case. Of course, it could be made configurable with many parameters, but in my opinion, it is simpler to do such logic in the app directly.

maciej-rexton:
If there is no similar API to the Stream based one (and after my small search through the docs there isn't) I totally agree with you that this API is much better. Thank you for explaining it.

sophisticode:
Rebased and ready for review again

sophisticode:
@boskokg: Is there any chance you will merge this PR to the base branch? Do you need more info?

gregk104:
Any word on this?  I'd really like to see this integrated as well.

boskokg:
@sophisticode can you please check the comment I put on your merge request?

android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -659,6 +659,24 @@ public void onMethodCall(@NonNull MethodCall call, @NonNull Result result) {
         break;
       }
 
+      case "readRssi":
```
sophisticode, line 662: Checked

README.md
```diff
@@ -162,6 +162,7 @@ For location permissions on iOS see more at: [https://developer.apple.com/docume
 | state                       |  :white_check_mark:  |  :white_check_mark:  | Stream of state changes for the Bluetooth Device. |
 | mtu                         |  :white_check_mark:  |  :white_check_mark:  | Stream of mtu size changes. |
 | requestMtu                  |  :white_check_mark:  |                      | Request to change the MTU for the device. |
+| readRssi                    |  :white_check_mark:  |  :white_check_mark:  | Read RSSI from a connected device. |
```
sophisticode, line 165: Check

sophisticode (commented):
check

sophisticode:
@boskokg: Sorry, but I cannot see your comment.

boskokg:
@sophisticode 
Screenshot 2022-04-04 at 13 43 02

sophisticode:
@boskokg:
The code in the example app was changed so that the stream stops when the device is disconnected.

boskokg:
Thanks @sophisticode

PR #20, opened 2022-03-04
add bluetoothDevice constructor from id

tanguypouriel:
My use case needed to take advantage of the fact that I already know the mac address of the device I want to talk to. This constructor method enable the ability to request to android to connect to a specific device as soon as it becomes available on the bluetooth "network".

### DIFF
--- a/example/pubspec.lock
+++ b/example/pubspec.lock
@@ -82,7 +82,7 @@ packages:
       path: ".."
       relative: true
     source: path
-    version: "1.1.1"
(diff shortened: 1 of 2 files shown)


boskokg:
Hello @tanguypouriel 

Your proposed code is commited with the small changes: 
* added documentation on the constructor, 
* setting the optional parameters for name and type
* skipping other changed files not related to mentioned proposal (merge conflicts...)

Thank you for the contribution.

Issue #31, opened 2022-03-24
Using the timeout parameter of BluetoothDevice.connect may cause two TimeoutExceptions to be thrown, one of which can't be caught


crazy-rodney:
The current implementation of BluetoothDevice.connect may result in two TimeoutExceptions when the timeout parameter is not null. Try running the following code on dartpad.dev:

```
import 'dart:async';

Future<void> test() async {
  final completer = Completer<void>();
  Timer? timer;
  timer = Timer(Duration(seconds: 1), () {
    completer.completeError(TimeoutException('Failed to connect in time.'));
  });
  await Future.delayed(Duration(seconds: 2));
  return completer.future;
}

void main() async {
  print('Trying to catch the exception:');
  try {
    await test();
  } catch (error) {
    print(error.toString());
  }
  
  print('Usin a guarded zone:');
  runZonedGuarded(
    test,
    (error, stackTrace) {
      print(error.toString());
    },
  );
}

```

As you see, the test function throws two errors, one of which can only be catched using a guarded zone.

The following would be a better approach to handling timeouts:

```
import 'dart:async';

Future<void> test() async {
  try {
    await Future(
      () async {
        // connect to device
      },
    ).timeout(const Duration(seconds: 2));
  } on TimeoutException catch (_) {
    // disconnect device
    throw TimeoutException('Failed to connect in time.', timeout);
  }
}

void main() async {
  try {
    await test();
  } catch (_) {
    print('Handling error');
  }
}
```

boskokg:
fixed in 1.2.0

PR #33, opened 2022-03-25
fix: uncatchable timeout exception in BluetoothDevice.connect

crazy-rodney:
#31


boskokg:
Hello @crazy-rodney.
Thank you for the contribution on this plugin.

I am not sure about the proposed change since it covers the mentioned issue, but what I see it also brings the change in the currrent code where we had the connection timeout optional feature (turned off by default), and now it is the mandatory feature (with the default timeout set to 20s).

Second thing is that anyone can wrap the connect method and do the timeout processing the way it suits the needs, so I am not sure if we need this issue fixed at all.

Can you please give some feedback on this?

sophisticode:
In my opinion, we should fix the timeout behaviour which does not work as intended (you cannot catch the exception within the app).

I propose to leave the timeout parameter optional and do something like this in the code: 

`      final connectFuture = Future(() async {
        var request = protos.ConnectRequest.create()
          ..remoteId = id.toString()
          ..androidAutoConnect = autoConnect;

        await FlutterBluePlus.instance._channel
            .invokeMethod('connect', request.writeToBuffer());

        await state.firstWhere((s) => s == BluetoothDeviceState.connected);
      });
      await (timeout != null ? connectFuture.timeout(timeout) : connectFuture);
`

SkuggaEdward:
I was about to make a PR for the same problem. Tho slightly different mod.
https://github.com/SkuggaEdward/flutter_blue_plus/blob/21d7bae2bb0b2fa5ba39d2cce2ad3c4adeb973f8/lib/src/bluetooth_device.dart#L42-L49

cthurston:
And one more way to solve it: https://github.com/Flowline/flutter_blue/pull/1/commits/de83c9ea06aabb9dbf5a85dbce760a659bbebf22

crazy-rodney:
I currently do not have time to spend on this problem. I've been using my solution since I proposed it and its working fine for me. Maybe merge mine or one of @SkuggaEdward's or @cthurston's. I think it would benefit the project to address this issue.

MousyBusiness:
```
Second thing is that anyone can wrap the connect method and do the timeout processing the way it suits the needs, so I am not sure if we need this issue fixed at all.
```

This is not true, you cannot catch the error as the timeout exception happens in a Timer handlers and doesn't propagate to the invoker. This is a major flutter_blue deficiency which has send me in search of a more active fork.

SimonH-Zuma:
I concur with all the above. I've been pulling my hair out wondering why I cannot catch the timeout in order to set my 'connected' flag accordingly. This is a MAJOR issue/bug so I would be grateful if you can release this PR

boskokg:
[SkuggaEdward](https://github.com/SkuggaEdward) proposal commited. closing merge request

PR #106, opened 2022-08-04
Fixed an error when calling the connect method of a connected device

figureai:
Fixed an error when calling the connect method of a connected device


figureai:
Fixed an error when calling the connect method of a connected device

Issue #66, opened 2022-05-23
The app stays on the "Bluetooth adapter is unknown" screen with iPhone (iOS 15.4.1)


sercan-uzun:
Hello folk. I am devoloping my first app on flutter with flutter_blue_plus.

If I run the application while it is already installed, I'm getting the "Bluetooth Adapter is unknown" error. If I run it again with the flutter run button or delete the application and reinstall it, the problem disappears. I don't see any error log on debug screen.
And It's not permissions issue. Because, if I restart with run button, working very well.

I also find that, this same problem has been experienced before, in the link below. It's look like a bug. How can I solve this problem? Thanks in advance.

https://github.com/pauldemarco/flutter_blue/issues/690

sercan-uzun:
By the way I am using example code. Calling Off screen like this:

class FlutterBlueApp extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      theme: ThemeData.dark(),
      color: Colors.lightBlue,
      home: StreamBuilder<BluetoothState>(
          stream: FlutterBluePlus.instance.state,
          initialData: BluetoothState.unknown,
          builder: (c, snapshot) {
            final state = snapshot.data;
            if (state == BluetoothState.on) {
              return const FindDevicesScreen();
            }
            return BluetoothOffScreen(state: state!); <-------- this line is calling Off screen with unknown.
          }),
    );
  }
}

arnsteinkjellevold:
I have the same issue, and I reported it before "flutter_blue_plus" was created.

I thought I found a workaround by accessing the "isOn" property before checking the state, however this only fixed it during development. Building a release to the users still had the same issue.

This issue started back in "flutter_blue" at the same time RXDart was updated from 0.26.X > 0.27.X so I suspect this is involved in the problem in some way.

arnsteinkjellevold:
My app is structured differently using a repository holding the bluettoth instance and a BLOC handling the bluetooth state so i'm not using streambuilder but I guess you could put something like this inside your build just before the return MaterialApp

    final _btInstance = FlutterBluePlus.instance;
    print(_btInstance.isOn.toString());
    sleep(const Duration(milliseconds: 100));

sercan-uzun:
> My app is structured differently using a repository holding the bluettoth instance and a BLOC handling the bluetooth state so i'm not using streambuilder but I guess you could put something like this inside your build just before the return MaterialApp
> 
> ```
[...]

PR #134, opened 2022-09-13
Fixes iOS apps getting stuck in unknown BT state

mbalamut:
Fixed an issue where iOS apps got stuck in the unknown Bluetooth state #66.
The cause of this problem was that after creating an CBCentralManager object, for a short while the state is unknown which is checked and reported by the plugin. Then two things happen in parallel:

- the plugin starts listening to the state stream,
- iOS calls centralManagerDidUpdateState with the real BT state.

If the centralManagerDidUpdateState call is faster, plugin doesn't get the first state update and is stuck in the unknown state.
Modified the code so that centralManagerDidUpdateState calls are cached before a stream event sink is attached.
I didn't see this problem on Android, but in theory it was also possible that state updates were skipped, so I modified the Android part to match changes for iOS.


xarmatura:
Your code duplicate BluetoothState on both platforms
I/flutter (10258): BluetoothState.off
I/flutter (10258): BluetoothState.off

chipweinberger:
I tested these changes on iOS, and they did fix https://github.com/boskokg/flutter_blue_plus/issues/66 for me.

The other changes look good as well.

Issue #135, opened 2022-09-15
Document the plugin's behavior of opting into neverForLocation on Android


navaronbracke:
I wanted to support BLE beacons when scanning for devices.
However the docs mention a caveat:

```
Note: If you include neverForLocation in your android:usesPermissionFlags,
some BLE beacons are filtered from the scan results.
```

The plugin's manifest specifically includes this flag.

I managed to fix that by adding an explicit entry to my Manifest.xml
```
<manifest xmlns:android="http://schemas.android.com/apk/res/android" xmlns:tools="http://schemas.android.com/tools"
    package="com.example.app"> <!-- needed for the tools annotation below -->

<!-- Override the permission in flutter_blue_plus's manifest with one without the permission flag -->
<uses-permission android:name="android.permission.BLUETOOTH_SCAN" tools:node="replace"/>
```
After adding this to my manifest, the flag is gone from my merged manifest, and the permission itself is still there.

I'd suggest we do one of the following:
1) Remove the `android:usesPermissionFlags="neverForLocation"` flag on the permission in the plugin's manifest
    and redirect people to [the Bluetooth docs](https://developer.android.com/guide/topics/connectivity/bluetooth/permissions)
    This is a breaking change (although not a big one)
2) Inform people that the plugin specifically opts into this flag and it can be removed by adding the permission explicitly with `tools:node="replace"`. The documentation on how this works is [here](https://developer.android.com/studio/build/manage-manifests)

This should fix #120 by providing people with either an opt-in or a clarification on how to _opt-out_, depending of what we  decide to do.

tuna-f1sh:
Thanks for this, saved me from making a fork just to fix this. I got caught by this as using the flutter_beacon plug-in in parallel. Particularly annoying since I read the comment in the example App manifest and thought, 'no, I don't want that flag' not knowing it was set within the plug-ins manifest.

Migration from flutter_blue original was painless other than this...

Issue #120, opened 2022-08-27
in Android 12 some beacon devices can not be found


erahani:
in Android 12 some beacon devices can not be found using this package.
I checked ScanResult device.name and it is empty for many scanned devices on Android 12. it works on iOS.
I also tried with some third-part apps like light blue and BLE scanner.  BLE scanner can not find it (while it can find same beacons in iOS app). light blue can find it on Android 12 and ios. I appreciate for any help.

navaronbracke:
The fact that some beacons can not be found on Android 12 is mentioned in the documentation here:
https://developer.android.com/guide/topics/connectivity/bluetooth/permissions#assert-never-for-location

```
Note: If you include neverForLocation in your android:usesPermissionFlags, some BLE beacons are filtered from the scan results.
```

The plugin does set `neverForLocation` in its manifest.

erahani:
Thanks for your reply:
i added this line but did not help. then I manually changed Location App permission to "Allow.." and it worked. not sure how I can avoid manual set of location.
<uses-permission android:name="android.permission.BLUETOOTH_SCAN"
        tools:remove="android:usesPermissionFlags"/>

PR #136, opened 2022-09-15
Opt out of neverForLocation on Android

navaronbracke:
This PR opts out of the `neverForLocation` permission flag on Android.

In general having to opt-in is better than having to explicitly opt-out.

Fixes: #135 Fixes: #120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 1.4.0
 **Breaking Change**
-* Opt-out of the `neverForLocation` permission flag for the `BLUETOOTH_SCAN` permission on Android.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 1.4.0
+**Breaking Change**
```
navaronbracke, line 2: @boskokg do we need an explicit link to

https://developer.android.com/guide/topics/connectivity/bluetooth/permissions

IIRC that is the only place where that flag is even mentioned and where people should be looking for an explanation of this change.

navaronbracke:
@boskokg is there a reason why version 1.4.0 was not published to pub.dev ?

PR #173, opened 2022-12-13
allow for simultaneous MAC and ServiceUuid ScanFilters

nairboon:
this way multiple different types of ScanFilters can be used


chipweinberger:
looks like a nice improvement.

PR #178, opened 2023-01-01
Fix stop scan api

### LINKED ISSUE #177: Missing getter for scanning state & stopScan() does not return early if not scanning
The bluetooth scanner Dart API does not expose the current state of the `isScanning` stream.
We only have access to a `Stream<bool>`, even though the underlying `StreamController` is a `BehaviorSubject`.

Secondly, the `stopScan()` method does not return early if the `_isScanning` controller indicates that the scanner is currently not scanning.

### Proposal

1. Provide a getter for the current scanning state, implemented like this:
```dart
  final BehaviorSubject<bool> _isScanning = BehaviorSubject.seeded(false);
  
  // Renamed, because the new getter can use the old name.
  Stream<bool> get isScanningStream => _isScanning; // BehaviorSubject implements the Stream interface, so no need to access the getter
  
  bool get isScanning => _isScanning.value;
```

2. Update the `stopScan()` method to return early if the scanner is not scanning, like this:
```
  /// Stops a scan for Bluetooth Low Energy devices
  Future<void> stopScan() async {
    if(!isScanning) {
      return;
    }
  
    await _channel.invokeMethod('stopScan');
    _stopScanPill.add(null);
    _isScanning.add(false);
  }
```

3. The `startScan()` method does not define the return type of the `Future`, it should be `Future<List<ScanResult>>`. Similarly, `stopScan()` should be `Future<void>`

navaronbracke:
Fixes #177

### DIFF
--- a/lib/src/flutter_blue_plus.dart
+++ b/lib/src/flutter_blue_plus.dart
@@ -195,7 +195,7 @@ class FlutterBluePlus {
   ///
   /// To observe the results while the scan is in progress, listen to the [scanResults] stream,
   /// or call [scan] instead.
-  Future startScan({
+  Future<List<ScanResult>> startScan({
     ScanMode scanMode = ScanMode.lowLatency,
     List<Guid> withServices = const [],
     List<Guid> withDevices = const [],
@@ -214,8 +214,8 @@ class FlutterBluePlus {
     return _scanResults.value;
   }
 
-  /// Stops a scan for Bluetooth Low Energy devices
-  Future stopScan() async {
+  /// Stops a scan for Bluetooth Low Energy devices.
+  Future<void> stopScan() async {
     await _channel.invokeMethod('stopScan');
     _stopScanPill.add(null);
     _isScanning.add(false);
--- a/lib/src/flutter_blue_plus.dart
+++ b/lib/src/flutter_blue_plus.dart
@@ -70,7 +70,12 @@ class FlutterBluePlus {
   }
 
   final BehaviorSubject<bool> _isScanning = BehaviorSubject.seeded(false);
-  Stream<bool> get isScanning => _isScanning.stream;
+
+  /// Returns whether the scanner is currently scanning.
+  bool get isScanning => _isScanning.value;
+
+  /// Returns the [Stream] of changes for the [isScanning] state.
+  Stream<bool> get isScanningStream => _isScanning;
 
   final BehaviorSubject<List<ScanResult>> _scanResults =
       BehaviorSubject.seeded([]);
--- a/lib/src/flutter_blue_plus.dart
+++ b/lib/src/flutter_blue_plus.dart
@@ -215,7 +215,13 @@ class FlutterBluePlus {
   }
 
   /// Stops a scan for Bluetooth Low Energy devices.
+  ///
+  /// Does nothing if currently not scanning.
   Future<void> stopScan() async {
+    if (!_isScanning.value) {
+      return;
+    }
+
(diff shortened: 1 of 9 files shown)


pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: flutter_blue_plus
 description: Flutter plugin for connecting and communicationg with Bluetooth Low Energy devices, on Android and iOS
-version: 1.4.0
+version: 1.4.1
```
ViliusP, line 3:
```suggestion
version: 1.5.0
```
Hi, I'm just randomly reviewing your PR :smile:. I think version should be bumped to 1.5.0 because you change public API by changing getter's `isScanning` type from `Stream<bool>` to  `bool` (in `lib/src/flutter_blue_plus.dart`). This is breaking change. Also, I think this change should be mentioned **emphasized** in `CHANGELOG.md`.

lib/src/flutter_blue_plus.dart
```diff
@@ -214,8 +219,14 @@ class FlutterBluePlus {
     return _scanResults.value;
   }
 
-  /// Stops a scan for Bluetooth Low Energy devices
-  Future stopScan() async {
+  /// Stops a scan for Bluetooth Low Energy devices.
+  ///
+  /// Does nothing if currently not scanning.
+  Future<void> stopScan() async {
+    if (!_isScanning.value) {
```
xarmatura, line 226: Hi, thanks for this PR. Just one additional note: to better know that this method failed, you should remove this check and wrap invokeMethod in a try-catch to handle the error, or throw an error on return here.

navaronbracke: That should not be necessary. If `await _channel.invokeMethod('stopScan');` fails, it'll throw a `PlatfomException` anyway.

xarmatura: Well, but on the other hand, how are we supposed to know that `stopScan` failed if it doesn't return anything in condition? Maybe it would be better to return `Future<bool>`?

navaronbracke: You just wrap it in a try catch on the consumer's side?

```dart
void myStopMethod() async {
  try {
    await FlutterBluePlus.instance.stopScan();
    
    print('Stop scan success');
  } on PlatformException catch (error) {
    print('stopScan failed');
  }
}
```

xarmatura: I meant here `if (!_isScanning.value) { return; }`.

navaronbracke: In that case, you'd check the new getter `isScanning` before calling `stopScan()`?

```dart
if(FlutterBluePlus.instance.isScanning) {
  await FlutterBluePlus.instance.stopScan();
}
```

xarmatura: Yeah, now looks pretty clear. Thanks for your time and sorry for my confusing question)

chipweinberger:
This looks good to me. If you can rebase, I can merge.

navaronbracke:
> This looks good to me. If you can rebase, I can merge.

Its been a while since I did the original changes. So I'd suggest you'd review the diff to see if I didn't mess it up.

chipweinberger:
Appreciate you looking at it. I think the rebase was not correct. it says "40 commits".

navaronbracke:
Yeah, I'll try again with a fresh branch. Should alleviate things a bit. I'll just cherry pick my commits there. Should be able to do that in a few hours from now.

PR #195, opened 2023-02-01
support for ScanFilters on Manufacturer data for Android

nairboon:
This allows scanning only devices that match the supplied filter on their Manufacturer data fields


chipweinberger:
Thanks for the PR.

To keep platform differences minimal, would love to merge if you could implement for iOS.

chipweinberger:
this code has significantly changed since removing protobufs. Will need to be updated.

Issue #198, opened 2023-02-06
Foreground Discover Services: no bluetooth gatt


ZainNomancowlar:
Im using the [flutter foreground task plugin](https://pub.dev/packages/flutter_foreground_task) and im getting some weird behaviour. The foreground service automatically connects to a particular device. In the Flutter/UI code I periodicaly call `FlutterBluePlus.instance.connectedDevices` to get the connected devices and the device shows up as a connected device. But when i try to get the sevices of the bluetooth device i get the error 
`discover_services_error, no instance of BluetoothGatt, have you connected first?`.  There is other wierd behaviour at times as well but I havent narrowed down when exactly that occurs.

lorenzosfarra:
Hi,

hope I am not off-topic here, but I have started from the same problem.
If it's off-topic, I will open a different issue for discussions/suggestions, but I think your opinion will not only assist me, but also benefit other developers who are seeking a reliable and solid strategy for managing foreground services and background logic related to BLE.

When it comes to foreground services and UI code, my main idea was to:

  - UI:  
    - getting a deviceId with `FlutterForegroundTask.getData<String>(key: 'deviceId')`. If present, basically do nothing
    - not present ? open a dialog to perform the scan of the devices, show them and save the deviceId when the user selects it with `FlutterForegroundTask.saveData(key: "deviceId", value: deviceId)`.
    
 - start the foreground service
     - getting a deviceId with `FlutterForegroundTask.getData<String>(key: 'deviceId')`. 
       - present? Connect to it and work with services and characteristics I already know (fixed IDs)
       - not present ? wait until the next onRepeatEvent to do the same

Thanks and have a nice day,
  Lorenzo

chipweinberger:
please try 1.7.6+, feel free to reopen if you still have problems.

PR #200, opened 2023-02-06
Fix for foreground isolate issue

ZainNomancowlar:
Fix for #198


chipweinberger:
Can you describe the fix more? What are these lines doing?

ZainNomancowlar:
I was having wierd issues when using the flutter blue plus with flutter_foreground_task plugin. for eg. If i started a bluetooth scan in the UI isolate, sometimes the scan results stream would get events in the foreground task but not in the UI isolate. I debugged things and found that two instances of the `FlutterBluePlusPlugin` class were being created. so i made some things static and made a map of method channels so that all data is sent to all isolates

xbug42:
I did not test this PR but making things static will not help if you have one FlutterBluePlus instance in each isolate. Isolates do not share memory. It's like separate processes for memory...
@ZainNomancowlar : doing scans in multiple isolates sounds like a good way to have all kinds of troubles to me... I really think only one isolate should handle all the bluetooth things....

ZainNomancowlar:
I havent got around to testing it much. I eventually decided to only use flutter blue plus in the foreground service. Plus i havent tested it at all in iOS. I think the changes might not be needed after all.

PR #201, opened 2023-02-08
Remove max sdk version

fewensa:
Some os must use these permissions, like xiaomi (Mi OS)


MrCsabaToth:
> Some os must use these permissions, like xiaomi (Mi OS)

I'm interested in that. What should I know about Xiaomi Mi OS related to these permissions? I want my app to work on as many devices as possible.

PR #204, opened 2023-02-10
Handling appearance standard advertisement data

### LINKED ISSUE #202: [Feature]: Support Appearance values
There are Categories and Subcategories defined here 
[Appearance Values.pdf](https://github.com/boskokg/flutter_blue_plus/files/10681601/Appearance.Values.pdf)

In some cases it's not enough to see what services a device exposes to exactly identify them. For example FTMS Fitness Machines hint their types (Indoor Bike, Treadmill, Rower, ...) via advertisement service data.

However in case of CSC (Cycling Speed and Cadence Sensors) there's no such service data, but I could tell apart speed and cadence sensors by looking at the appearance. That would hint the type. I saw that nRF Connect can sense that appearance without connecting to a device and polling the characteristics.

Bellow is an nRF scan, this time with a HRM:
Screenshot_2023-02-07-20-29-46-77_3c2e989be58abe9ca7463ffb1e6967d2

MrCsabaToth:
Fixes #202
Not tested with iOS, I wrote the code blindly.
I tested it with Android, but for some reason the data is null, just like all the Tx data as well. nRF Connect displays the appearance though. So I don't know if this is some problem with my development environment or the code.

### DIFF
--- a/android/src/main/java/com/boskokg/flutter_blue_plus/AdvertisementParser.java
+++ b/android/src/main/java/com/boskokg/flutter_blue_plus/AdvertisementParser.java
@@ -87,6 +87,10 @@ class AdvertisementParser {
           ret.setTxPowerLevel(Protos.Int32Value.newBuilder().setValue(data.get()));
           break;
         }
+        case 0x19: { // Appearance
+          ret.setAppearance(Protos.Int32Value.newBuilder().setValue(data.getShort()));
+          break;
+        }
         case 0x16: // Service Data with 16 bit UUID.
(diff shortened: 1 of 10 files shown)


chipweinberger:
Thanks for the PR.

closing due to: "Not tested with iOS, I wrote the code blindly."

Feel free to re-open the PR if you have a chance to test on iOS.

MrCsabaToth:
That's not something I'll be able to do in the near future due to the walled garden nature of Apple. I wish we could build and test an iOS build on Windows or Linux...

PR #239, opened 2023-04-18
[Dependencies] Remove unnecessary dependencies: convert, collection, meta

chipweinberger:
Related Issue: https://github.com/boskokg/flutter_blue_plus/issues/236

Remove:
- convert
- collection
- meta

With this simple change, flutter_blue_plus will only have 3 dependencies:
- flutter
- rxdart
- protobuf

There are big advantages to having few dependencies. Namely, you avoid:
- a new version of a dependency causing bugs
- incompatibility between versions
- explosion of child dependencies
- unnecessary library complexity 

I understand these dependencies are some of the safest to include, but they are are incredibly simple to remove.

### DIFF
--- a/lib/flutter_blue_plus.dart
+++ b/lib/flutter_blue_plus.dart
@@ -7,8 +7,6 @@ library flutter_blue_plus;
 import 'dart:async';
 import 'dart:io';
 
-import 'package:collection/collection.dart';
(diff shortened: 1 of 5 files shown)


WeiKuang:
very good

chipweinberger:
Tested on iOS & Merged.

PR #240, opened 2023-04-18
[Dependencies] Remove rxdart dependency

chipweinberger:
Related Issue: https://github.com/boskokg/flutter_blue_plus/issues/236

**Note: this needs more testing! Draft PR.**

 I've started work on removing RxDart. Please just view this commit: https://github.com/boskokg/flutter_blue_plus/pull/240/commits/3cb0a5ec36d7afcb80c017135efe6e9c0483c67f

**Motivation**: RxDart is rather complicated, often updated, and rather overkill. It is only used in a few places for what seem like trivial reasons but it's implementation is hard to understand using multiple layers of abstraction.

There are big advantages to having few dependencies. Namely, you avoid:

- a new version of a dependency causing bugs
- incompatibility between versions
- explosion of child dependencies
- unnecessary library complexity
- binary size

### DIFF
--- a/lib/flutter_blue_plus.dart
+++ b/lib/flutter_blue_plus.dart
@@ -9,7 +9,6 @@ import 'dart:io';
 
 import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
-import 'package:rxdart/rxdart.dart';
 
 import 'gen/flutterblueplus.pb.dart' as protos;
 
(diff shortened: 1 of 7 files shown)


WeiKuang:
very good.you are very smart.
The community needs people like you.

chipweinberger:
Tested working in the iOS example.

PR #243, opened 2023-04-20
[Dependencies] Remove protobuf, replace with OpenAPI

chipweinberger:
**This is a draft.** 

**Motivation:**
- the main goal is to lower the maintenance burden of flutter_blue_plus
- the secondary goal is to make flutter_blue_plus simpler to read and understand
- protobuf is a complicated runtime dependency
- openAPI generates idiomatic JSON based serialization code with few dependencies

**Results:**
- the generated dart code is great, much better than protobuf
- the generated objc code is great, much better than protobuf
- the generated java code is meh. worse.

**Todo:**
- still need to wire things up, if we were to go down the approach. This PR is just to showcase the generated code.


WeiKuang:
I agree with you.
You have done a great job, look forward to your other performances

chipweinberger:
closing. too complicated. not a clear win.

chipweinberger:
I should have used https://pub.dev/packages/pigeon

PR #253, opened 2023-04-27
Port macos support from flutter_blue with clang-format

tuna-f1sh:
Re: https://github.com/boskokg/flutter_blue_plus/pull/140

PR #269, opened 2023-05-02
bare minimum macOS support

tuna-f1sh:
#253 with upstream changes and non-working methods commented out.

I've checked and works as per previous integration. It's not ideal as you've found out with most of this library! 

If better support is desired, I believe the auto-generated in 'macos/gen/' should be re-generated. I tried copying the current 'ios/gen' but I got protobuf version errors - even though the pod version is the same? I don't have the time at the moment to go any deeper into this and don't have setup to do the auto-generated files.


chipweinberger:
Yes this looks good.

chipweinberger:
Would you be able to try the regeneration? Looks like it would only take 5 minutes.

Screenshot 2023-05-02 at 2 02 00 PM

chipweinberger:
if you run into more issues, no worry. I'll just merge.

tuna-f1sh:
I should have been clearer; the problem is not generating the code but that the generated code doesn't build. The macOS build complains that `protoc` version mismatches the generated files and I'm not sure why - it has the same version dependency as iOS.

I actually had to figure out the generation for https://github.com/boskokg/flutter_blue_plus/pull/270 . The file in https://github.com/boskokg/flutter_blue_plus/blob/master/protos/regenerate.md is more useful than that ChatGPT reply since it has the context.

It should be updated however:
1. Install protoc plugin globally `dart pub global activate protoc_plugin` [ref](https://pub.dev/packages/protoc_plugin)
2. Add pub bin to path `PATH="$HOME/.pub-cache/bin:$PATH"`
3. Run the following commands from this project's protos folder protoc --dart_out=../lib/gen ./flutterblueplus.proto protoc --objc_out=../ios/gen ./flutterblueplus.proto

tuna-f1sh:
Ok I took the time to resolve this too. I had to run `pod update` in the project I'm using to pull in the Protobuf version update in https://github.com/boskokg/flutter_blue_plus/pull/269/commits/b1dce7db81e87bae88017c5e0747183e2aef1066#diff-02c41fdb3f24915419275601c0a4772a79baf3fd13c9699a6741240730435f56R26

Looking at this though, I believe https://github.com/boskokg/flutter_blue_plus/blob/master/ios/flutter_blue_plus.podspec#L26 should also be updated to reflect the `protoc` version used to generate the files.

chipweinberger:
Do you think it's ready to merge? let me know!

tuna-f1sh:
Yes, I've now updated that and the regenerate.md file. Should be good to merge now.

chipweinberger:
merging.

PR #270, opened 2023-05-02
[Android] support set preferred phy

tuna-f1sh:
Referring to https://developer.android.com/reference/android/bluetooth/BluetoothGatt#setPreferredPhy(int,%20int,%20int)

Also I had to add a { at https://github.com/boskokg/flutter_blue_plus/compare/master...tuna-f1sh:set-preferred-phy?expand=1#diff-e49ebc6c8561aa50689fd6b03f2927d12578c1defbc53f18a26010bde8506afeR824 for current to compile on Android


chipweinberger:
looks good to me. Ready to merge?

tuna-f1sh:
Yes it is.

PR #273, opened 2023-05-03
scan timeout timer property so can be cancelled

tuna-f1sh:
The current method of calling stopScan on timeout causes a rogue Future if `stopScan()` if called whilst scanning, that causes erroneous stops.

For example:

1. call `startScan(timeout: 20)`
2. call after 10 seconds `stopScan()`.
3. call `startScan(timeout: 20)`
4. scan will stop due to first timeout Future after 10 seconds not 20.
5. call `startScan()`.
6. scan will stop after 10 seconds due to second call timeout even though no timeout desired.

This adds a Timer property to the class such that `stopScan()` can cancel it and prevent this behaviour.


chipweinberger:
nice job spotting this. thanks!

PR #324, opened 2023-06-06
Close BufferStream listen on stopScan

ma2yama:
When I ran `startScan` again after `stopScan`, I got data twice.


chipweinberger:
thanks.

PR #354, opened 2023-06-27
Bug Fixed - allowDuplicates not actually used on Android

GrayLand119:
Bug Fixed:
- allowDuplicates not actually used on Android


chipweinberger:
thanks

PR #355, opened 2023-06-27
[Android] Regression at 75b07df fixed. Change MTU did not work.

bbajzi:


chipweinberger:
looks good except please remove the support for Lollipop. 

Lollipop was released in 2014. There is no need to support it.

bbajzi:
> looks good except please remove the support for Lollipop.
> 
> Lollipop was released in 2014. There is no need to support it.

The code does not support Lollipop. Requesting mtu will return error on Lollipop. This is the purpose of the enclosing "if" statement. This "if" statement could be removed and gatt.requestMtu could be called unconditionally (the removed gatt.requestMtu was a regression). But minSdkVersion 19 should be increased in that case. I don't want to increase it if not necessary. I have customers using proprietary tablets with old versions of Android (e.g. Decent Espresso). At that products my app does not call requestMtu and no error will occur.

bbajzi:
I have checked again. Increasing minSdkVersion to 21 enables removing the version checking code. (But please, don't increase more until not forced :-) )

chipweinberger:
merged 
https://github.com/boskokg/flutter_blue_plus/commit/c3ac392148805cf9293b7b2812ba603b277193d4
https://github.com/boskokg/flutter_blue_plus/commit/267427b372e0610a02c20c8614e9f99a8ccad057

PR #364, opened 2023-07-08
rename former "from() calls" to new versions

Staars:
This is not heavily tested, but seems to go into the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 ## Foreward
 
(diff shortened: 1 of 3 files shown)


Staars:
Already fixed here:
https://github.com/boskokg/flutter_blue_plus/commit/5067b5b75a0a1087db4029a67f79fe98f96a8949

Issue #370, opened 2023-07-15
ACCESS_FINE_LOCATION - PlatformException - Version 1.7.1 - Android 13


cedar-yxs:
When I run example app, then I click search button, I had an error
Logs as below:
```
E/flutter ( 5612): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(startScan, FlutterBluePlus requires android.permission.ACCESS_FINE_LOCATION permission, null, null)
E/flutter ( 5612): #0      StandardMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:653:7)
E/flutter ( 5612): #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:315:18)
E/flutter ( 5612): <asynchronous suspension>
E/flutter ( 5612): #2      FlutterBluePlus.scan (package:flutter_blue_plus/src/flutter_blue_plus.dart:189:7)
E/flutter ( 5612): <asynchronous suspension>
```

I run it with flutter 3.7.12 and 3.10.6. I had the same error

The action button is 'TURN OFF'. I hit the button and nothing happened

chipweinberger:
perhaps we/you need to update the permissions here

https://github.com/boskokg/flutter_blue_plus/blob/master/example/android/app/src/main/AndroidManifest.xml

huayun502:
me too,I update the permissions,but not working - Android 13

chipweinberger:
it may be due to this

https://github.com/boskokg/flutter_blue_plus/commit/c0b1de652b141a5962477b48d6cd6de757fa36f0

perhaps you can add breakpoints and debug if the permission is being requested

huayun502:
E/flutter (15423): [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: PlatformException(startScan, FlutterBluePlus requires android.permission.ACCESS_FINE_LOCATION permission, null, null)
E/flutter (15423): #0      StandardMethodCodec.decodeEnvelope
message_codecs.dart:652
E/flutter (15423): #1      MethodChannel._invokeMethod
platform_channel.dart:310
E/flutter (15423): <asynchronous suspension>
E/flutter (15423): #2      FlutterBluePlus.scan
flutter_blue_plus.dart:189
E/flutter (15423): <asynchronous suspension>

chipweinberger:
[...]

PR #377, opened 2023-07-17
#370 fix Android 13 access fine location error

FritzMatthaeus:
This will fix the Android 13 Error

```
FlutterBluePlus requires android.permission.ACCESS_FINE_LOCATION permission
```

ACCESS_FINE_LOCATION was restricted to maxSDK 30 in Plugin. This is not overwritten by App Manifest, therefore it will crash on Android 13. So i have removed that restriction.

Example and ReadME have been updated.


chipweinberger:
thanks

Issue #334, opened 2023-06-11
BLE Scan Continuously 


Kavinkumarsimple:
How do i continuously scan for BLE Devices? Im building a BLE Scanner app and it should always scan and show the changes

GrayLand119:
> How do i continuously scan for BLE Devices? Im building a BLE Scanner app and it should always scan and show the changes

Try this branch, I created a commit to support scan continuously.
Set params updateInterval, expiredInterval, and allowDuplicates=true in scan method.

https://github.com/GrayLand119/flutter_blue_plus/commit/5b5afa145cf7a19103e44e8f8f62d0a77fb346ae

R-Emagina:
Hi,

I'd love to see this in the master branch.
Does someone know when it could be merged?

Thanks!

GrayLand119:
PR Created: https://github.com/boskokg/flutter_blue_plus/pull/382

chipweinberger:
thanks for the PR, but AFAICT, it should be easy for the user to implement this feature themself. 

Just create a timer and repeatedly call `FlutterBluePlus.instance.scan` yourself, etc.

This repo already has a ton of bugs. We can't handle more complexity.

Thanks again.

PR #382, opened 2023-07-19
New Feature - Scan continuously

GrayLand119:
Fixed issue:
[BLE Scan Continuously #334 ](https://github.com/boskokg/flutter_blue_plus/issues/334)

Added:
- Add new params(updateInterval, expiredInterval) with scan. 

Updated:
- Update example to scan continuously

Set updateInterval, expiredInterval, and allowDuplicates=true, for fixing situations that may cause an error:
1. Discovered a bluetooth device and it in scan results.
2. Power-off the device to simulate the device have been leaved scene.
3. The device still in discovered list(scan results). So user can connect the device, but will never get success.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -180,12 +180,22 @@ class FindDevicesScreen extends StatelessWidget {
               backgroundColor: Colors.red,
             );
           } else {
+            /// previously method
+            // return FloatingActionButton(
+            //     child: const Icon(Icons.search),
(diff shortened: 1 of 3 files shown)


chipweinberger:
thanks for the PR, but AFAICT, it should be easy for the user to implement this feature themself.

This repo already has a ton of bugs. We can't handle more complexity.

Thanks again.

PR #387, opened 2023-07-19
[Android] fix JSON success value is boolean

bma73:
The `WriteCharacteristicResponse` returns a boolean as the success value, not a number. Therefore, the check `json['success'] != 0` will always evaluate to true, even if the peripheral responded with an error.

```
WriteCharacteristicResponse:
 {request={service_uuid=dd3f469f-10c2-403f-94b0-79f46a288d00, remote_id=6A:44:8B:60:D8:35, characteristic_uuid=dd3f469f-10c2-403f-94b0-79f46a288d0a, write_type=0, value=}, success=false}
```

### DIFF
--- a/lib/src/bluetooth_msgs.dart
+++ b/lib/src/bluetooth_msgs.dart
@@ -570,7 +570,7 @@ class BmWriteDescriptorResponse {
     _printDbg("\nBmWriteDescriptorResponse $json");
     return BmWriteDescriptorResponse(
       request: BmWriteDescriptorRequest.fromMap(json['request']),
-      success: json['success'] != 0,
+      success: json['success'],
     );
   }
 }


MrCsabaToth:
This looks to be a good find. However if we fix `BmWriteCharacteristicResponse` (https://github.com/boskokg/flutter_blue_plus/blob/23d86e125db84aa30febcf00d4c25e0c7eac474b/lib/src/bluetooth_msgs.dart#L505)
then we should fix `BmWriteDescriptorResponse` (https://github.com/boskokg/flutter_blue_plus/blob/23d86e125db84aa30febcf00d4c25e0c7eac474b/lib/src/bluetooth_msgs.dart#L571) and `BmSetNotificationResponse` too (https://github.com/boskokg/flutter_blue_plus/blob/23d86e125db84aa30febcf00d4c25e0c7eac474b/lib/src/bluetooth_msgs.dart#L623)

chipweinberger:
this is intentional because Objective-C code does not support booleans as map values. 

instead we should have the android code return integers, not booleans.

chipweinberger:
fixed: afbdaa62c886107a3cea51d2f674c4a356345f26

PR #396, opened 2023-07-20
[iOS] Crash on Connect when reconnecting to existing device

RubenKid:
I can reproduce this issue in the latest version (1.8.2) https://github.com/boskokg/flutter_blue_plus/issues/343 

Fixed it catching a Flutter Error and also storing the known peripherals in a list as strong reference

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 1.8.7
+* dart: add 15 seconds default timeout for ble communication  
+
 ## 1.8.6
 * dart: rename BluetoothDevice.id -> remoteId
 * dart: rename uuid -> characteristicUuid / serviceUuid / descriptorUuid
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -1,6 +1,6 @@
 name: flutter_blue_plus
 description: Flutter plugin for connecting and communicationg with Bluetooth Low Energy devices, on Android and iOS
-version: 1.8.6
+version: 1.8.7
 homepage: https://github.com/boskokg/flutter_blue_plus
 
 environment:


kalismeras61:
@RubenKid when using the app, the app crashes when you switch bluetooth off. Is this fix for it ? 

`FBP-iOS] handleMethodCall: setNotification
*** Terminating app due to uncaught exception of class 'FlutterError'
libc++abi: terminating due to uncaught exception of type FlutterError
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00000001df58d158 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`:
->  0x1df58d158 <+8>:  b.lo   0x1df58d174               ; <+36>
    0x1df58d15c <+12>: stp    x29, x30, [sp, #-0x10]!
    0x1df58d160 <+16>: mov    x29, sp
    0x1df58d164 <+20>: bl     0x1df588b30               ; cerror_nocancel
Target 0: (Runner) stopped.`

kalismeras61:
@RubenKid i tried your branch but still same issue when you switch off ble app crash.

chipweinberger:
I don't see what this fixes. What is the crash trace that this fixes?

RubenKid:
@kalismeras61 @chipweinberger  my crash didnt happen in that situation. It was happening when you connected a device before so you have its remoteId and in later launch of the app you generate the BluetoothDevice object using Proto and try to connect to it. The findPeripheral method was throwing a FlutterError but the code is catching a NSException so it crashed. After that the app didnt crash but didnt connect either so I needed to add the device to a strong reference list

chipweinberger:
thanks for the update. the first part makes sense, but no longer needed (please rebase).

I don't see why you would need a separate strong reference here.

PR #400, opened 2023-07-21
Keep known peripherals in strong reference Array

RubenKid:
@chipweinberger closed the other PR because the crash didn't happen anymore although still need to keep the strong reference to known device to avoid this error

[CoreBluetooth] API MISUSE: Cancelling connection for unused peripheral <CBPeripheral: 0x28320e700, identifier = <remote_id>, name = <name>, mtu = 0, state = connecting>, Did you forget to keep a reference to it?

### DIFF
--- a/ios/Classes/FlutterBluePlusPlugin.m
+++ b/ios/Classes/FlutterBluePlusPlugin.m
@@ -36,6 +36,7 @@ typedef NS_ENUM(NSUInteger, LogLevel) {
 @property(nonatomic, retain) FlutterBluePlusStreamHandler *stateStreamHandler;
 @property(nonatomic, retain) CBCentralManager *centralManager;
 @property(nonatomic) NSMutableDictionary *scannedPeripherals;
+@property(nonatomic) NSMutableDictionary *knownPeripherals;
 @property(nonatomic) NSMutableArray *servicesThatNeedDiscovered;
 @property(nonatomic) NSMutableArray *characteristicsThatNeedDiscovered;
 @property(nonatomic) NSMutableDictionary *dataWaitingToWriteWithoutResponse;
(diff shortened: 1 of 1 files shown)


chipweinberger:
This warning message in iOS CoreBluetooth appears when you have a `CBPeripheral` that you are attempting to connect or are connected to, but then you lose all references to it in your application code. When the `CBPeripheral` object is deallocated while the connection is still ongoing, iOS considers it a misuse of the API.

This situation could happen if, for example, you store the `CBPeripheral` object in a local variable within a function, and then the function ends, causing the `CBPeripheral` object to be deallocated. To avoid this, you should maintain a strong reference to the `CBPeripheral` object for as long as you need the connection – usually as an instance variable in your class that's handling the Bluetooth interactions.

For example:

```swift
class MyBluetoothManager: NSObject, CBCentralManagerDelegate {
    var centralManager: CBCentralManager!
    var connectedPeripheral: CBPeripheral!  // <-- maintain a strong reference to your peripheral

    // rest of your class...
}
```

Then, when you're done with the peripheral (either you've disconnected, or you're done with the connection), you can set `connectedPeripheral` to `nil` to let it be deallocated if you wish.

chipweinberger:
That is what ChatGPT says^^

seems like we should remove the strong reference when disconnect.

also please add a comment why we need this string reference. it is confusing without a comment.

chipweinberger:
also, which reference is being lost?

Where in the code are we storing weak references to peripherals?

Without answering this question we should not merge a 'fix'.

RubenKid:
I am not Obj-C expert but I saw this behaviour when you have a deviceId to connect to but left the app and come back again (so you didnt connect in that session). When try to connect to that device somehow I am getting that error and seems it loses the reference. We could store just the connected device but I thought could be useful storing it in an array but both solutions should work

chipweinberger:
Thanks for the update. I'll try and reproduce the error. If I see it, I will properly fix it.

chipweinberger:
I've researched the problem and submitted 1.10.5 as a fix.

PR #416, opened 2023-07-23
iOS, MacOS: fixed ScanResult.advertisementData.serviceUuids

ekuleshov:
Fix for https://github.com/boskokg/flutter_blue_plus/issues/415

It looks like a typo in `ios/Classes/FlutterBluePlusPlugin.m` for both iOS and MacOS.


chipweinberger:
fixed in 1.9.5 thanks for submitting

PR #444, opened 2023-07-27
ignore MTU on characteristic write

FritzMatthaeus:
I ran into a problem with the added MTU Size check in the 'writeCharacteristic' method.

I develop an App for one specific device and the MTU Size i get from the device is the minimum size 23. I know from the hardware developer that i can and should send a larger byte array to speed up data transer. Negotiating a different MTU size by requesting it keeps failing. 

So i came up with a "ignoreMtuRestriction" flag in the write method to explicitely ignore the MTU size as i know for sure that the receiving client can handle that byte array size. 
```
await c.write([0x12, 0x13], ignoreMtuRestriction: true);
```

I've tested it on Android & iOS. I also added this (very special) case to the readMe in the Debugging section.

unfortunately my IDE keeps re-formatting your code, so a lot of format changes, sorry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@
 
 FlutterBluePlus is a Bluetooth Low Energy plugin for [Flutter](https://flutter.dev).
 
-**Bluetooth Classic is not supported.** 
+**Bluetooth Classic is not supported.**
 
 (i.e. speakers, headphones, mice, keyboards, gamepads, and more are not supported).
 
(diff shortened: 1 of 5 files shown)


chipweinberger:
you just need to set your formatter line limit to 120 characters.

I would probably call this "mtuCheck" and default it to true.

chipweinberger:
for now though i don't think i want to merge this without thinking about it more.

chipweinberger:
Maybe characteristic.setAdvancedOptions({bool mtuCheck = true)

chipweinberger:
i've pushed to master with a fix for this.

chipweinberger:
I may revert this.

Issue #453, opened 2023-07-31
[Feature]: Remove "print" from write-Function


aklehm:
### Feature Request

Please remove the `print("withoutResponse $withoutResponse");` from the write-Function in the `/lib/src/bluetooth_characteristic.dart` file.
It is very annoying while sending bigger files. If you still want the print statement for debug, you should add a debug flag to that function.

chipweinberger:
thanks. it's a mistake

chipweinberger:
fixed e53b1e1261064f6fc6063d3ec074a5dcbc740d98

will be in `1.12.9`

chipweinberger:
1.12.9 is now released

PR #455, opened 2023-07-31
fix: use `debugPrint` instead of `print` for logging information

dhruvanbhalara:
- fixes https://github.com/boskokg/flutter_blue_plus/issues/453

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -518,7 +518,7 @@ class DeviceScreen extends StatelessWidget {
       try {
         yield await device.readRssi();
       } catch (e) {
-        print("Error reading RSSI: $e");
+        debugPrint("Error reading RSSI: $e");
         break;
       }
       await Future.delayed(frequency);
--- a/lib/flutter_blue_plus.dart
+++ b/lib/flutter_blue_plus.dart
@@ -7,6 +7,7 @@ library flutter_blue_plus;
 import 'dart:async';
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 
 part 'src/bluetooth_characteristic.dart';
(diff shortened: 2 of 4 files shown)


MrCsabaToth:
That's a great find. It's also great that you are importing the foundations and not the cupertino (comes first choice in ABC as the possible choices besides foundations and material)

dhruvanbhalara:
@MrCsabaToth @chipweinberger I've used `debugPrint` for reason, if app is overriding `debugPrint` as per application/developer requirements in that case configuration is applied to it. in other case if `print` is used that will be use own confirmation and it's visible in production build as well.

chipweinberger:
@dhruvanbhalara , thanks for the PR. 

I've already pushed a fix for #453 

There is not need for `debugPrint` because logs can be turned off with `FlutterBluePlus.setLogLeve(LogLevel.none)`

MrCsabaToth:
https://github.com/boskokg/flutter_blue_plus/commit/e53b1e1261064f6fc6063d3ec074a5dcbc740d98 looks like on `print` only from `write`. This PR modified 3 more places where there are (were?) `print`.

chipweinberger:
we should keep those as 'print'. 

`debugPrint` can skip logs to increase perf, which is not what we want. To increase perf, users should set 
`setLogLevel`.

I know dart likes to complain about print, but there is no issue to fix. I've already disabled the warnings for that.

MrCsabaToth:
As long as the `print` is not executed in release version then OK. I didn't know `debugPrint` can be skipped for performance, I'll read about that.

benthillerkus:
> debugPrint can skip logs to increase perf, which is not what we want, because it hurts debugging.

AFAIK it's actually the opposite: [Android may drop lines with `print`, but throttles the log output with `debugPrint`, so that nothing is printed](https://arc.net/l/quote/bqutvrky).

PR #460, opened 2023-08-02
[iOS]fix CBDescriptor.vaule type conversion errors

Temoa:
When I read the device descriptor, it returns an NSString object, which only returns [0,0,0,0] in dart. So I modified it here, if it is NSString, convert it to NSData and return it.


chipweinberger:
Screenshot 2023-08-02 at 12 13 08 AM

chipweinberger:
Thanks for the PR! 

Looks like we should handle NSNumber as well.

chipweinberger:
merged: https://github.com/boskokg/flutter_blue_plus/commit/36da1ac1ff4d92fee8cb18e4afc52741e4bbfdce

Temoa:
Thank you for your work.

PR #500, opened 2023-08-15
[iOS] fix: manufacturer data not parsed if only company ID is present

keep-cloud-tr:
Manufacturer data may only include the company ID (16 bits). We still want to parse it in this case.

See also:
https://github.com/pauldemarco/flutter_blue/pull/293/commits/bd6071fcd78ecdff61a6f64d3546aafb090cc0bf


chipweinberger:
nice pr =) And nice job following the commit message convention!

PR #537, opened 2023-09-05
[Android]: Handle autoConnect.

MartinXBcn:
Android: In case of autoConnect=true do not close Gatt whenn a connected device disconnects, but keep connection-data for a later re-connect. Only close Gatt when disconnect was initiated by the user.

Unfortunately I have not found a way to read the 'autoConnect' state once set; that is why I added a new map  'mAutoConnect' (like the maps 'mConnectedDevices' and 'mMtu') to store the autoConnect-state.

### DIFF
--- a/android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -96,6 +96,7 @@ public class FlutterBluePlusPlugin implements
 
     private final Map<String, BluetoothGatt> mConnectedDevices = new ConcurrentHashMap<>();
     private final Map<String, Integer> mMtu = new ConcurrentHashMap<>();
+    private final Map<String, Boolean> mAutoConnect = new ConcurrentHashMap<>();
 
     private int lastEventId = 1452;
     private final Map<Integer, OperationOnPermission> operationsOnPermission = new HashMap<>();
@@ -471,6 +472,10 @@ public class FlutterBluePlusPlugin implements
                         if (gatt == null) {
                             result.error("connect", String.format("device.connectGatt returned null"), null);
                             return;
+                        } else {
+                            // Save flag "autoConnect"; 
+                            // if connect was not succesful entry will be deleted in Gatt-callback.
+                            mAutoConnect.put(remoteId, autoConnect);
                         }
 
                         result.success(0);
@@ -481,6 +486,9 @@ public class FlutterBluePlusPlugin implements
                 case "disconnect":
                 {
                     String remoteId = (String) call.arguments;
+
+                    mAutoConnect.remove(remoteId);
+                    // Entry in mConnectedDevices and mMtu will be removed in the corresponding "Gatt callback".
 
                     // already disconnected?
                     BluetoothGatt gatt = mConnectedDevices.get(remoteId);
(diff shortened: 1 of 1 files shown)


android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -1424,24 +1433,42 @@ public void onConnectionStateChange(BluetoothGatt gatt, int status, int newState
 
             String remoteId = gatt.getDevice().getAddress();
 
+            if (status != BluetoothGatt.GATT_SUCCESS) {
+                Boolean autoConnect = mAutoConnect.get(remoteId);
+                if ((autoConnect != null) && (autoConnect == false)) {
+                    // Connect failed and no autoConnect set, delete entry in mAutoConnect again which was just set
+                    // in "Method call" "call".
+                    mAutoConnect.remove(remoteId);
```
chipweinberger, line 1441: do we need to remove this? what would happen if we just leave the map as-is

android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -471,6 +472,10 @@ public void onMethodCall(@NonNull MethodCall call,
                         if (gatt == null) {
                             result.error("connect", String.format("device.connectGatt returned null"), null);
                             return;
+                        } else {
+                            // Save flag "autoConnect"; 
+                            // if connect was not succesful entry will be deleted in Gatt-callback.
+                            mAutoConnect.put(remoteId, autoConnect);
```
chipweinberger, line 478: maybe we should do this earlier?

like right after we get the argument?

chipweinberger:
does it fix autoconnect?

thanks

MartinXBcn:
> does it fix autoconnect?
> 
> thanks

Yes, this fixes the autoConnect-issue under Android: if ``autoConnect`` was set to ``true`` calling ``connectGatt(...)`` then in case of a disconnect caused by the connected device (i.e. out-of-range) ``gatt.close()`` is not called anymore; ``gatt.close()`` is called only in case that the flutter-app actively runs "disconnect".

android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
MartinXBcn: The idea is to wait for the result of the call of ``connectGatt(...)`` and if it the call itself was successful to avoid create an entry in ``mAutoConnect`` unnecessarily.
If the call itself was successful but later the connection failed then the created entry in ``mAutoConnect`` will be deleted in the gatt-callback.

ericaig:
I can also confirm that this fixes the autoConnect issue I was also having 👌🏾

chipweinberger:
I've pushed this fix, inspired by yours. 

3d3bd1063cd778d17d2b0dec3e57a1f11d3dd5fe

android/src/main/java/com/boskokg/flutter_blue_plus/FlutterBluePlusPlugin.java
chipweinberger: unecessary entries in `mAutoConnect` does not bother me

PR #577, opened 2023-09-19
Shiny lagoon added get method

ShinyLagoon:


chipweinberger:
just create a class extension, you can access this with an extension

PR #611, opened 2023-10-09
Hot fix for crash on iOS while discovering services

tonimoeckel:
There was a crashed forced by an unknown selector: [_NSInlineData intValue]: unrecognized selector sent to instance 

I fixed this by adding a check for the class type. 

A merge would be appreciated. 

This is related to: https://github.com/pauldemarco/flutter_blue/issues/1092

Best,
Toni


chipweinberger:
btw, data is actually unused here and should just be deleted =)

tonimoeckel:
I was wondering about this . but unsure ^^

PR #627, opened 2023-10-16
Android: "getAdapterName" requires bluetooth-permissions

MartinXBcn:
"getAdapterName" needs already bluetooth-permissions, at least under Android; if not any other function ("connect", "turnon" etc.) which requests permission is called before, call of "getAdapterName" ends in an exception.

Therefore "ensurePermissions"-code added before calling mBluetoothAdapter.getName().


chipweinberger:
looks good but please use better commit message and use a single commit

MartinXBcn:
I joined the two commits to one single commit with a new title.

chipweinberger:
sorry i should have been more specific

the title should be

[Fix] get adapterName must request permission"

MartinXBcn:
Done.

chipweinberger:
thanks!

PR #628, opened 2023-10-16
Manager to handle multiple devices and streams

manuel75o:
Hey, I created a BluetoothManager to handle multiple devices and their streams.
I tested it with latest version of FBP on IOS17.
Hope you consider Pulling it. I am happy about feedback.

Best regards Manuel

### DIFF
--- a/lib/src/bluetooth_manager.dart
+++ b/lib/src/bluetooth_manager.dart
@@ -1,0 +1,154 @@
+// Copyright 2023, Manuel Reischer.
+// All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+// BluetoothManager's aim is to handle multiple devices and their Streams
+
+part of flutter_blue_plus;
+
+/// DeviceModel represents the device and its corresponding Services and Characteristics
+/// A device is unique to a DeviceModel, There can be multiple Services and Characteristics per DeviceModel
+class DeviceModel {
+  final BluetoothDevice device;
+  final List<BluetoothService> services;
+  final List<CharacteristicStream> characteristicsStream;
+  DeviceModel(this.device, this.services, this.characteristicsStream);
+}
+
+/// CharacteristicsStream contains the BluetoothCharacteristic and Info wether it's Stream is initialized
+class CharacteristicStream {
+  final BluetoothCharacteristic characteristic;
+  bool isInitialized;
+  CharacteristicStream({
+    required this.characteristic,
+    this.isInitialized = false,
+  });
+}
+
+/// BluetoothManager's aim is to handle multiple devices and their Streams
+class BluetoothManager {
+  final deviceModel = <DeviceModel>[];
+  int deviceCount = 0;
+
+  /// Connects a Device and checks if the Device is already connected
+  /// Check by remoteId if the device is already connected
+  /// Connect the Device if not connected
+  /// Calls "discoverDeviceServiceAndCharacteristic"
+  void connectDevice(BluetoothDevice device) async {
+    bool isDeviceAlreadyConnected = deviceModel.any((deviceModel) => deviceModel.device.remoteId == device.remoteId);
+    if (!isDeviceAlreadyConnected) {
+      await device.connect();
+      discoverBluetoothServiceAndCharacteristic(device);
+    } else {
+      print("Device already connected!");
+    }
+  }
+
+  /// Disconnects a Device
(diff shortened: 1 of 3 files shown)


chipweinberger:
thanks for the PR

this is the kind of layer that should be build on top of FBP as a separate project

otherwise it is confusing why we have two ways to do everything

you should consider releasing it yourself with a separate name. perhaps `BluetoothManager`

that said to be honest I don't fully understand the need for it :) You can already do all of this with the BluetoothDevice class

lib/src/bluetooth_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+
+  /// UTF8-Decodes the received Data and returns it
+  String dataParser(List<int> dataFromDevice) {
+    return utf8.decode(dataFromDevice);
+  }
+
+  /// Sends Data to the Characteristic
+  /// UTF8-Encodes the Data
+  void writeCharacteristicData(String characteristicData, BluetoothCharacteristic characteristic) async {
+    List<int> bytes = utf8.encode(characteristicData);
```
chipweinberger, line 138: not all characteristics want utf8

manuel75o:
@chipweinberger got ya, thanks for the feedback.
I tried to address the problem many people have when it comes to handling multiple devices & streams. Since FBP is obviously capable of many things.
When it comes to this topic some functionality's are missing.

Such as having a general Structure to store such devices/services/streams etc.

And second thing being able to access the characteristic/streams with ease.

Nonetheless i will continue work on my BluetoothManager lib.
Best regards :)

lib/src/bluetooth_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+    await characteristic.write(bytes);
+  }
+
+  /// Checks if the device is connected if it loses the connection by accident
+  /// it removes the DeviceModel to counteract
+  checkDeviceConnection() {
+    for (var dm in deviceModel) {
+      final device = dm.device;
+      device.connectionState.asBroadcastStream().listen((event) {
```
chipweinberger, line 147: this should be setup only during initialization.

chipweinberger:
i added some comments 😄 

i think it can use refinement.

manuel75o:
@chipweinberger i totally understand the decision.Thank you a lot for the feedback. May I stay in touch with you?

chipweinberger:
in 1.19.2, I also added

`device.cancelWhenDisconnected(subscription)`

this makes it easier to manage devices & streams using vanilla FlutterBluePlus

lib/src/bluetooth_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+    return deviceModel;
+  }
+
+  /// Returns a Characteristic by device and CharacteristicNumber
+  /// Since it is possible to have multiple Characteristic per Service, you can get the right Characteristic
+  /// by device and it's CHaracteristicNumber (Starting at 0)!
+  BluetoothCharacteristic getCharacteristic(BluetoothDevice device, int characteristicNumber) {
+    final dm = deviceModel.firstWhere((deviceModel) => deviceModel.device == device);
+    final characteristic = dm.characteristicsStream[characteristicNumber].characteristic;
+    return characteristic;
```
chipweinberger, line 111: characteristics will have different "numbers" on iOS & android. 

FlutterBluePlus does not guarantee a specific order of characteristics.

lib/src/bluetooth_manager.dart
```diff
@@ -0,0 +1,154 @@
...
+  final BluetoothDevice device;
+  final List<BluetoothService> services;
+  final List<CharacteristicStream> characteristicsStream;
+  DeviceModel(this.device, this.services, this.characteristicsStream);
+}
+
+/// CharacteristicsStream contains the BluetoothCharacteristic and Info wether it's Stream is initialized
+class CharacteristicStream {
+  final BluetoothCharacteristic characteristic;
+  bool isInitialized;
```
chipweinberger, line 21: isInitialized is not used anywher?

manuel75o:
@chipweinberger i like the changes so far, thank you

lib/src/bluetooth_manager.dart
manuel75o: @chipweinberger its more like an index than "number". Since the characteristicStream is within a list you need to know at which index you characteristic is.
Thats straightforward, because you decide the the order when programming your ble device (esp32) etc.

PR #636, opened 2023-10-21
Fixed a migration error in 1.20.3.

manuel75o:
new function: Introduced a new `connectedDevices`. Caution: a previous function used this same name. That older function is now called `systemDevices`.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -190,3 +190,6 @@ BluetoothDevice? myDevice = await myDeviceFuture;
 ## 1.20.0
 
 * **renamed:** `connectedSystemDevices` -> `systemDevices`, because they must be re-connected by *your* app.
+
+## 1.20.3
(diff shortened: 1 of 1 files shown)


chipweinberger:
this is more confusing than just not mentioning it imo.

manuel75o:
@chipweinberger nontheless it states an error, which was fixed by this.

chipweinberger:
the migration guide already mentions that breaking change in 1.11.0, and 1.20.0.

yes it is kind of confusing. but there is no breaking change in 1.20.3

chipweinberger:
`* **new function:** Introduced a new `connectedDevices`. **Note**: a previous function had this same name, but that older function is now called `systemDevices`. `

If you use this description, I'll add it =)

manuel75o:
@chipweinberger will do ;)

manuel75o:
@chipweinberger like this? or the title?

PR #637, opened 2023-10-21
Fix: added missing migration text for 1.20.3

manuel75o:
@chipweinberger added your suggested migration text. If you want you can delete the old PR.


chipweinberger:
looks good.

the commit title should be 

"[Docs] added migration caution for 1.20.3"

you can change the title using

```
git commit --amend -m "..."
git push force
```

PR #649, opened 2023-11-01
feat : add get bytes

shbae-musma:
get bytes 구현


chipweinberger:
this will not work on iOS. Therefore, I do not want to add it.

Why do you need it?

shbae-musma:
I will make sure to modify the iOS source code later. Thank you very much.

chipweinberger:
iOS does not allow raw bytes.

It cannot be supported on iOS.

shbae-musma:
Wow, I didn't know that, thank you for letting me know!

PR #658, opened 2023-11-04
fixed perpetual scan

Ibetz1:
Here is a branch containing my quick fix the the bug I posted.


chipweinberger:
fixed it in 1.24.1

PR #673, opened 2023-11-10
[Android] Compare both short versions of UUID in getDescriptorFromArray

RubenKid:
When looking for descriptor we are trying to compare the short version of the UUID with the long one so if never found. Sorting both of them


chipweinberger:
thanks. similar fix is in master branch now.

PR #691, opened 2023-11-19
Added matching of serviceUuid during characteristic search

Romcik077:
Issue appear when I tried to crate descriptors for my characteristics with same uuid but under different services with different uuid. As result I had descriptor for all characteristics but only from first one, even if characteristic has another descriptor with another uuid or doesn't have any descriptors. With next code everything works as expected.


chipweinberger:
youre right this is a bug. thanks!

PR #716, opened 2023-12-05
fix connect method args

utamori:
In the connect method, it's casting to a String, 
but since what's being passed is a Map, it's throw error

https://github.com/boskokg/flutter_blue_plus/blob/a2d86d0f3038356525d005e0297e2491735b9bd8/lib/src/bluetooth_device.dart#L117

https://github.com/boskokg/flutter_blue_plus/blob/a2d86d0f3038356525d005e0297e2491735b9bd8/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java#L664

### DIFF
--- a/lib/src/bluetooth_device.dart
+++ b/lib/src/bluetooth_device.dart
@@ -85,8 +85,8 @@ class BluetoothDevice {
     int? mtu = 512,
     bool autoConnect = false,
   }) async {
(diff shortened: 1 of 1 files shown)


chipweinberger:
already fixed. thanks.

Issue #758, opened 2024-01-17
[Feature]: cancelWhenDisconnected function for connectionState listener


spehj:
### FlutterBluePlus Version

1.31.8

### Flutter Version

3.16.5

### What OS?

All

### OS Version

Android 13

### Bluetooth Module

Laird BL652-SA

### What is your feature request?

Would it be possible to modify a cancelWhenDisconnected to let the last BluetoothConnectionState.disconnected be emitted before the subscription is canceled?

It would be useful to track the connection state for each device, do something when the device is disconnected, and then clear the subscription.

Here is a simple example for illustration:

```dart
// listen for disconnection
var subscription = device.connectionState.listen((BluetoothConnectionState state) async {
        print("$state");
// Right now it prints BluetoothConnectionState.disconnected, then BluetoothConnectionState.connected on connect. 
// It doesn't print anything on disconnect, since the subscription is probably canceled before the last connection state is emitted.
});

// ADDED this: Cancel subscription when state is disconnected
device.cancelWhenDisconnected(subscription, next:true);

// Connect to the device
await device.connect();

// Disconnect from device
await device.disconnect();

// cancel to prevent duplicate listeners
subscription.cancel();
```

### Logs

```shell
No logs
```

chipweinberger:
ive thought about this before.

feel free to open a PR

spehj:
Yes, if `_methodStream.add(call);` was moved I think would probably save this case. 

For now, I will just cancel the subscription inside the `connectionState` listener.

Minimal example for anyone looking for this (note that when you connect for the first time, it emits `disconnected` value).

```dart

bool isConnected = false; // Use this to not cancel the stream if not yet connected
StreamSubscription<BluetoothConnectionState>? subscription;
      
// listen for disconnection
subscription = device.connectionState.listen((BluetoothConnectionState state) async {
        print("$state");
        // Check if the state is 'disconnected'
    if (!isConnected && state == BluetoothConnectionState.disconnected) {
        // Don't do anything before BluetoothConnectionState.connected
    } else if (state == BluetoothConnectionState.disconnected){
        // Cancel the subscription if it's 'disconnected'
        await subscription?.cancel();
        subscription = null; // Clear the reference
        print("Disconnected and subscription canceled");
    } else if (state == BluetoothConnectionState.connected) {
        bool isConnected = true;
    }

});

// Don't need this
// device.cancelWhenDisconnected(subscription, next:true);

// Connect to the device
await device.connect();

[...]

PR #763, opened 2024-01-20
`BluetoothConnectionState.disconnected` emitted before subscription is canceled when using `device.cancelWhenDisconnected` method

spehj:
This PR fixes the `BluetoothConnectionState.disconnected` not being emitted before `device.connectionState` subscription is canceled when using `device.cancelWhenDisconnected(subscription, next: true);`.

See issue #758.


chipweinberger:
support added in 1.31.9

PR #773, opened 2024-01-24
Fix type 'DeviceIdentifier' is not a subtype of type 'String'

a1573595:
Fix connectedDevices type cast exception.


chipweinberger:
thanks

chipweinberger:
fixed 1.31.10

PR #800, opened 2024-02-29
Added nearest device feature in android

ganeshshirole:
Added nearest device feature in android


chipweinberger:
sorry. lots of bad changes here
- kotlin is unecessary 
- min sdk needs to be 21, not 26
- you shouldnt change the java version 
- increased the number of android files by 5x 
- tons of unnessary formatting changes
- afaict there is no reason this should be implemented in the android code. It should be in Dart code

PR #873, opened 2024-05-06
Merging multiple MSDs with the same manufacturer ID

MrCsabaToth:
Fixing https://github.com/boskokg/flutter_blue_plus/issues/785
1. Needed to change input parameters for proposed `getManufacturerSpecificData`
2. Also changed from `List<Integer>` to `byte[]`, because `getBytes` returns that, and in the other end `bytesToHex` requires that as well. So it's unnecessary weight lifting to to convert to List<Integer>. We just merge `byte[]` chunks with `System.arraycopy`.

### DIFF
--- a/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -1758,20 +1758,19 @@ public class FlutterBluePlusPlugin implements
         int n = 0;
 
         while (n < bytes.length) {
-
-            int fieldLen = bytes[n];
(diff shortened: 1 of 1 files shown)


MrCsabaToth:
This is a proposal

MrCsabaToth:
Note that I also made parts of the `getAppearanceFromScanRecord` function use the similar constructs as the newly introduced and proposed function.

chipweinberger:
what api level is entrySet?

chipweinberger:
I've reverted for now until we can verify iOS behavior

MrCsabaToth:
Let's disregard my old iPad, @zwells confirmed that it's consistent on Android vs iOS, we'll possible be even more consistent on various devices than nRF!

Issue #927, opened 2024-07-05
[Help]: Android: Race condition between disconnect and connect leaves connection stranded


jasaw:
### Requirements

- [X] I've looked at the README 'Common Problems' section

### Have you checked this problem on the example app?

No

### FlutterBluePlus Version

1.32.8

### Flutter Version

3.22.2

### What OS?

Android

### OS Version

7 & 14

### Bluetooth Module

Nordic nRF

### What is your problem?

Sometimes calling `bluetoothDevice.connect(autoConnect: false)` then shortly after (around 500ms), calling `bluetoothDevice.disconnect(queue: false)` does not cancel the connect call, but FBP thinks that there's no active Bluetooth connection. `bluetoothDevice.isDisconnected` returns true. This leaves the active Bluetooth connection stranded because calling `bluetoothDevice.disconnect(queue: false)` does nothing.

Sometimes FBP throws this exception `FlutterBluePlusException | connect | fbp-code: 10 | connection canceled`, sometimes it doesn't, but regardless of the exception, the outcome is the same, stranded Bluetooth connection.

### Logs

```shell
D/[FBP-Android](32473): [FBP] onMethodCall: connect
D/BluetoothGatt(32473): connect() - device: DA:71:BA:6D:29:3B, auto: false
D/BluetoothAdapter(32473): isSecureModeEnabled
D/BluetoothGatt(32473): registerApp()
D/BluetoothGatt(32473): registerApp() - UUID=bcf24f3c-294a-473c-abe9-c78b490a8d23
D/BluetoothGatt(32473): onClientRegistered() - status=0 clientIf=7
I/flutter (32473): 0:01:06.344230	I	Device:	 Disconnecting Bluetooth connection
D/[FBP-Android](32473): [FBP] onMethodCall: disconnect
I/flutter (32473): 0:01:06.346947	D	Device:	 BLE device disconnected
D/[FBP-Android](32473): [FBP] disconnect: cancelling connection in progress
D/BluetoothGatt(32473): cancelOpen() - device: DA:71:BA:6D:29:3B
I/flutter (32473): 0:01:06.347598	W	Device:	 Bluetooth device isDisconnected: true
D/BluetoothGatt(32473): close()
D/BluetoothGatt(32473): unregisterApp() - mClientIf=7
I/flutter (32473): 0:01:06.419830	W	Device:	 Disconnecting Bluetooth connection again
D/[FBP-Android](32473): [FBP] onMethodCall: disconnect
[...]

PR #933, opened 2024-07-16
Android: enforce 2 second disconnect delay

jasaw:
Addresses issue #927.
This workaround enforces a 2 second delay between connect and disconnect.
Android bug report: https://issuetracker.google.com/issues/37121040

### DIFF
--- a/lib/src/bluetooth_device.dart
+++ b/lib/src/bluetooth_device.dart
@@ -185,7 +185,15 @@ class BluetoothDevice {
   ///   - [queue] If true, this disconnect request will be executed after all other operations complete.
   ///     If false, this disconnect request will be executed right now, i.e. skipping to the front
   ///     of the fbp operation queue, which is useful to cancel an in-progress connection attempt.
-  Future<void> disconnect({int timeout = 35, bool queue = true}) async {
+  ///   - [androidDelay] Android only. Minimum gap in milliseconds between connect and disconnect to
+  ///     workaround a race condition that leaves connection stranded.
+  ///     https://issuetracker.google.com/issues/37121040
+  ///     From testing, 2 second delay appears to be enough.
+  Future<void> disconnect({
+    int timeout = 35,
+    bool queue = true,
+    int androidDelay = 2000,
+  }) async {
     // Only allow a single disconnect operation at a time
     _Mutex dtx = _MutexFactory.getMutexForKey("disconnect");
     await dtx.take();
@@ -211,7 +219,7 @@ class BluetoothDevice {
       Future<BmConnectionStateResponse> futureState = responseStream.first;
 
       // Workaround Android race condition: ensure minimum connect disconnect gap is met
-      await _ensureMinimumAndroidDelay();
+      await _ensureMinimumAndroidDelay(androidDelay);
 
       // invoke
       bool changed = await FlutterBluePlus._invokeMethod('disconnect', remoteId.str);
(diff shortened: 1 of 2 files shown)


lib/src/bluetooth_device.dart
```diff
@@ -203,13 +210,34 @@ class BluetoothDevice {
...
+        if (FlutterBluePlus._connectTimestamp.containsKey(remoteId)) {
+          const Duration minGap = Duration(milliseconds: 2000);
+          Duration elapsed = DateTime.now().difference(FlutterBluePlus._connectTimestamp[remoteId]!);
+          if (elapsed.compareTo(minGap) < 0) {
+            Duration timeLeft = minGap - elapsed;
+            print("[FBP] disconnect: enforcing ${minGap.inMilliseconds}ms disconnect gap, delaying ${timeLeft.inMilliseconds}ms");
+            await Future<void>.delayed(timeLeft);
+          }
+        }
+      }
```
chipweinberger, line 227: can you move to separate function

_ensureMinimumAndroidDelay()

chipweinberger:
the delay should be configurable, i think.

In FlutterBluePlus.setOptions

lib/src/bluetooth_device.dart
jasaw: Done. Workaround is contained in `_ensureMinimumAndroidDelay` function.

jasaw:
Updated the PR. Delay is configurable, defaults to 2s.

chipweinberger:
thanks for the PR and being receptive to changes

thinking about the api more

it should probably be better as an argument to the disconnect function 

and called `androidDelay`

thoughts?

jasaw:
>it should probably be better as an argument to the disconnect function

I imagine most people wouldn't mess with the default 2s delay unless they want to go through vigorous testing themselves, so hiding it in `FlutterBluePlus.setOptions` makes more sense.

chipweinberger:
appreciate your thoughts

I think with a good a good function comment it will be clear why it is there. 

but yes, i think it is better as a disconnect argument. 

our api already has precendent for this (see requestMtu)

jasaw:
Yup, done, moved to disconnect argument. 👍

chipweinberger:
looks good. I'd call it 

_ensureAndroidDisconnectionDelay

chipweinberger:
I'd also simplify this comment since we already have the comment elsewhere

      // record connection time
      if (Platform.isAndroid) {
        FlutterBluePlus._connectTimestamp[remoteId] = DateTime.now();
      }

chipweinberger:
Also I dont think we need this code. seems unecessary to remove it I think.

      if (Platform.isAndroid) {
        // Disconnected, remove connect timestamp
        FlutterBluePlus._connectTimestamp.remove(remoteId);
      }

chipweinberger:
also "workaround a race condition that leaves connection stranded."

id clarify what "leaves connection stranded." means

chipweinberger:
also, im still not convinced we need both this 2-second delay change + the other PR.

The other PR seems sufficient. Why is it not sufficient?

jasaw:
>_ensureAndroidDisconnectionDelay

Done

>simplify this comment since we already have the comment elsewhere

Done

>Also I dont think we need this code. seems unecessary to remove it I think, and makes things more complicated

I don't like ever growing memory usage. Yes it's just a small amount of memory, but my app is designed to run for at least 3 months without interruption or restart, and it scans and makes Bluetooth connection while running, all without human intervention. This ever growing memory usage is just going to look exactly like a memory leak when I profile the memory usage.

>id clarify what "leaves connection stranded." means

Done.

>also, im still not convinced we need both this 2-second delay change + the other PR.
The other PR seems sufficient. Why is it not sufficient?

This 2-second delay is to workaround an Android bug that establishes connection even though gatt has been disconnected and closed. On successful connection, FBP's gatt callback function is **not** called, so we end up with a connection that cannot be killed because there is no gatt.

chipweinberger:
4015129121bfcb22f68ffbda758e34225e29f9d8

Issue #948, opened 2024-07-25
[Feature]: add support for `includedServices` (i.e. secondary services)


DHT-Xavier:
### Requirements

- [X] I've looked at the README 'Common Problems' section

### Have you checked this problem on the example app?

Yes

### FlutterBluePlus Version

1.32.11

### Flutter Version

3.22.3

### What OS?

Android

### OS Version

13

### Bluetooth Module

Build-in BLE module on Samsung Galaxy Tab A9+

### What is your problem?

I am using FBP to connect a weight scale using BLE. The scale has 2 service (`0x181D` for weight and `0x181B` for body composition). Each of them should contain 2 characteristics for `READ` and `INDICATE` with a Descriptor for CCC.

Characteristics `0x2A9E`, `0x2A9D` under service `0x181D` are working fine
However, Characteristics `0x2A9B`, `0x2A9C` under service `0x181B` (which `isPrimary: false`) are not able to `READ` or `INDICATE`

A `PlatformException` return while tried to `setNotifyValue(true)` on `0x2A9C`
```
I/flutter (12268): [FBP] <setNotifyValue> args: {remote_id: 00:09:1F:82:D1:50, 
service_uuid: 181d, secondary_service_uuid: null, characteristic_uuid: 2a9c, 
force_indications: false, enable: true}
D/[FBP-Android](12268): [FBP] onMethodCall: setNotifyValue
I/flutter (12268): PlatformException(setNotifyValue, characteristic not found in 
service (chr: '2a9c' svc: '181d'), null, null)
```
> Please note that the error mentions `characteristic not found in 
service (chr: '2a9c' svc: '181d')`, `0x2A9C` should 'belong' to service `0x181B` but not `0x181D`

Seems while I am trying to do anything on the characteristic under service which is not primary service, it tried to sending command to it's serviceUuid but not the secondaryServiceUuid. To temporary fix this issue, I change the code in 
[`flutter_blue_plus-1.32.11\lib\src\bluetooth_characteristic.dart` line `248`](https://github.com/boskokg/flutter_blue_plus/blob/54e81550204156127f20c719feae5b98d435bb5f/lib/src/bluetooth_characteristic.dart#L248)

from 
```
serviceUuid: serviceUuid,
```
to
```
serviceUuid: secondaryServiceUuid ?? serviceUuid,
```
So that I can temporary `setNotifyValue` to the characteristic

Here are some details of the `isPrimary: false` service and characteristic under the service.

[...]

PR #858, opened 2024-04-19
Adding windows

jurapusic:
I added windows part of  flutter_blue_plugin into flutter_blue_plus.
I need to test it.

PR #971, opened 2024-08-18
refactor: migrate to federated architecture

tnc1997:
https://github.com/boskokg/flutter_blue_plus/pull/969#issuecomment-2294775575


tnc1997:
Hi @chipweinberger, I have been manually testing the Android implementation, however I do not have access to an iOS nor macOS device, would you be able to assist with the manual testing of the iOS and macOS implementation?

chipweinberger:
yes I'll try to test it

chipweinberger:
hi, looking good. Just some pre-review feedback.

1. I dont like the models being scattered around the file structure so much. Before they were all in the same file -- I think thats best. Or at least all in the same folder.
2. please don't add any dependencies. This increases maintenance cost. the only dependency we need to add is "plugin_platform_interface: ^2.0.0"

I appreciate you adding serialization tests, but I think its of little value. It increases maintenance cost, code base, and serialization bugs are not hard to find.

tnc1997:
> Or at least all in the same folder.

I have moved all types into the same folder which aligns with the convention of the Flutter [plugins](https://github.com/flutter/packages) for consistency.

> If its possible to avoid using plugin_platform_interface, I would prefer that too.

I am pretty sure that [`plugin_platform_interface`](https://pub.dev/packages/plugin_platform_interface) is fundamental but @stuartmorgan may be able to confirm.

> No dependencies is a feature of FBP.

I have removed the two dependencies in question and replaced them with the implementations as previously.

> I appreciate you adding serialization tests, but I think its of little value.

I personally believe that automated tests covering logic (e.g. serialization and deserialization above and beyond simply `'property': property` and `property: json['property']` respectively or method channel communications) improve maintainability and increase confidence. I wholeheartedly agree that automated tests are unnecessary and of little to no value where no logic is taking place, therefore I have not added any automated tests for these occurrences. Automated tests are reusable, and once developed, they can be reused across multiple development cycles with little to no additional cost. The platform interface is highly likely to remain largely stable, therefore I do not believe that there is a significant ongoing maintenance overhead in the long-term. They can be advantageous for commonly used packages with a high number of dependants, such as this one, that require ongoing maintenance and support, and they are more accurate and much faster than manual testing, because they eliminate the possibility of human error and can be run either locally or within a CI/CD environment in a matter of seconds. With automated tests in the platform interface, we would be able to confidently and safely focus manual testing efforts on the Android/iOS/macOS implementations themselves, without having to worry about unintentional breakages or changes to behaviour in the platform interface.

stuartmorgan:
> I am pretty sure that [`plugin_platform_interface`](https://pub.dev/packages/plugin_platform_interface) is fundamental but @stuartmorgan may be able to confirm.

Just putting the new `base` keyword on the platform interface class will prevent the same mistake that `plugin_platform_interface` is designed to prevent (see https://github.com/flutter/flutter/issues/127396).

packages/flutter_blue_plus_platform_interface/lib/src/method_channel/method_channel_flutter_blue_plus.dart
stuartmorgan, line 1: FWIW, I would recommend against putting a method channel implementation in your platform interface package unless there are existing, published, third-party implementations that were written by relying on the internal details of the method channel that you don't want to break. (This is the only reason flutter.dev plugins tend to have this pattern; there were a number of relatively high-profile implementations of packages built that way prior to the existence of federation.)

If the method channel is in the platform interface package, then the entire serialization surface becomes a de-facto API, and you can never make any breaking changes to the method channel itself without the very disruptive step of making a major version change to the platform interface.

All of the flutter.dev plugins treat that as a legacy implementation that is never used in our own implementations, and in newer plugin implementations we don't have it at all. All channel communication is internal to each implementation package. If the goal is just to share this implementation between Android and iOS, you could just put the Android and iOS implementations in the same package. Alternately, you could do what flutter.dev plugins do and have a separate copy of serialization in each platform implementation. That allows changing the details of the channel without having to keep N different endpoints in sync. E.g., if you need to add a parameter to a method channel method only for Android (in such a way that the arguments change structure, like from a single object to a dictionary), you can change it in Dart and Java, without having to make sure you also remember to change Objective-C code to match to avoid crashes.

tnc1997:
> Just putting the new `base` keyword on the platform interface class will prevent the same mistake that `plugin_platform_interface` is designed to prevent (see [flutter/flutter#127396](https://github.com/flutter/flutter/issues/127396)).

Apologies, I had not realised that `plugin_platform_interface` was deprecated, and have refactored accordingly.

packages/flutter_blue_plus_platform_interface/lib/src/method_channel/method_channel_flutter_blue_plus.dart
tnc1997: Thank you very much for your review, you raise some very good points. Would you recommend basing the platform interface on the functionality that is available in the public static methods of [`flutter_blue_plus.dart`](https://github.com/boskokg/flutter_blue_plus/blob/master/lib/src/flutter_blue_plus.dart) and refactoring that accordingly instead of the existing method channel?

stuartmorgan: I don't know anything about this specific plugin, so I don't have any basis to recommend anything about what the actual platform interface should be.

tnc1997: > That allows changing the details of the channel without having to keep N different endpoints in sync.

That sounds like something that may be very useful in this case especially given the differences between the Bluetooth implementations on Android vs iOS/macOS.

chipweinberger:
@tnc1997 

but why do we need to refactor the model file structure and add tests as part of this change to federated?

this PR is already huge. I'd like to keep it just the minimally required changes for federated.

tnc1997:
> I'd like to keep it just the minimally required changes for federated.

I presume therefore that you would prefer to use the existing method channel methods as the platform interface?

> add tests as part of this change to federated?

I am happy to remove the automated tests that cover the logic, I thought that these may be useful in increasing our confidence of the changes.

Out of interest would you be willing to consider reviewing these separately as a follow-up change in order to make these changes easier to review?

chipweinberger:
@tnc1997 , I added you as a collaborator :) 

> I thought that these may be useful in increasing our confidence of the changes.

Did the serialization code change? Maybe I dont understand how federated works.

chipweinberger:
> I presume therefore that you would prefer to use the existing method channel methods as the platform interface?

That was my original idea yes, but afaik it wont work with the Web platform implementation.

chipweinberger:
> flutter_lints: ^4.0.0

do we need this? I bothers me that they are on version 4.0.0 already. When they release 5.0.0, are we going to need to update? I'd rather have less maintanence.

chipweinberger:
and, given how disruptive this change seems to be, I still think `flutter_blue_universal` seems like a good idea.

Why don't you like that idea?

chipweinberger:
the more I think about it, the more I think a new package `flutter_blue_universal` makes sense.

`flutter_blue_plus` is very stable. It's in maintenance mode.

Adding web, linux, windows, these are big changes, and the platform interface will probably need to be revised multiple times. 

I dont think this active development is suitable for `flutter_blue_plus`.

tnc1997:
> Did the serialization code change? Maybe I dont understand how federated works.

No, the serialization logic has not changed, the files have simply moved into the platform interface.

> The only feature I care about is letting other users implement their own platform implementations.

Creating a platform interface allows other third parties to create non-endorsed implementations of this package.

There were no tests covering the logic previously, see https://github.com/chipweinberger/flutter_blue_plus/pull/971#issuecomment-2298146942 for some benefits of automated testing.

> For example, look at this: [PhilipsHue/flutter_reactive_ble#836](https://github.com/PhilipsHue/flutter_reactive_ble/issues/836). It's an issue asking to update an internal dependency. I don't want anything like that for FBP.

It appears that the issue that you linked to is referring to a `dependency` and not a [`dev_dependency`](https://dart.dev/tools/pub/dependencies#dev-dependencies).

> Dev dependencies differ from regular dependencies in that dev dependencies of packages you depend on are ignored. Pub gets every package that your package depends on, and everything those packages depend on, transitively. It also gets your package's dev dependencies, but it ignores the dev dependencies of any dependent packages. Pub only gets your package's dev dependencies.

Regardless of the above I am happy to remove the development dependency if you would prefer.

tnc1997:
> That was my original idea yes, but afaik it wont work with the Web platform implementation you made.

Other third party implementations should set the `instance` property to their particular implementation.

For example a Linux implementation should set the `instance` property to e.g. `FlutterBluePlusLinux`.

```dart
class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
  /// Registers this class as the default instance of [FlutterBluePlusPlatform].
  static void registerWith(Registrar registrar) {
    FlutterBluePlusPlatform.instance = FlutterBluePlusLinux();
  }

  // TODO: override supported methods here
}
```

> Also, your linux implementation was implemented in Dart, not native.

Defaulting to the method channel implementation means that no changes are required to Android/iOS/macOS.

chipweinberger:
Personally - I think we should make the interface at the dart layer, and have duplicated dart code.

chipweinberger:
RE: `flutter_blue_universal`

If we could make this PR much smaller, I would be more comfortable merging it into `flutter_blue_plus`.

chipweinberger:
Another thing to consider, each platform has it's own specific features.

for example, on android we have `requestConnectionPriority`, but not on any other platform.

When web, or windows, or android adds new methods, it would be ideal if the platform interface does not need to change.

tnc1997:
> When web, or windows, or android adds new methods, it would be ideal if the platform interface does not need to change.

With this in mind how would you envisage a common platform interface shared between platforms?

Should we provide a minimal platform interface with only the functionality that is common across all platforms?

tnc1997:
> The biggest decision is where to make the platform interface, i.e. dart or native.

Based on this should we instead base the platform interface on the public static methods in [`flutter_blue_plus.dart`](https://github.com/boskokg/flutter_blue_plus/blob/master/lib/src/flutter_blue_plus.dart)?

chipweinberger:
> Based on this should we instead base the platform interface on the public static methods in [`flutter_blue_plus.dart`](https://github.com/boskokg/flutter_blue_plus/blob/master/lib/src/flutter_blue_plus.dart)?

I'm leaning towards this, yes. Seems more future proof.

tnc1997:
> I think we would want something like this:
> 
> `FlutterBluePlus.android.requestConnectionPriority(....)`
> 
> And the maintainer of Android can add their own functions anytime.

Apologies if I am misunderstanding but that sounds like we want a minimal truly common platform interface and then consumers can depend on platform implementations if they would like to call platform specific methods?

chipweinberger:
> Apologies if I am misunderstanding but that sounds like we want a minimal truly common platform interface and then consumers can depend on platform implementations if they would like to call platform specific methods?

Yes, I think that seems smart.

I don't know enough about Linux, Windows, Web to know if its truly a good idea though.

This is why I think we should do development in some other less-stable package. There's a lot we might want to break.

tnc1997:
> Yes, I think that seems smart.
> 
> I don't know enough about Linux, Windows, Web to know if its truly a good idea though.

It seems like a reasonable idea. Any consumers who are only interested in using cross-platform functionality would depend on only `flutter_blue_plus` as today. Whereas consumers who are interested in using cross-platform and/or platform specific functionality would depend on that platform's specific implementation e.g. `flutter_blue_plus_android` which would contain the platform interface implementation and the platform specific functionality.

chipweinberger:
At a higher level, I just want to reiterate, **my biases are minimal maintenance and long term stability.** 

So, I'll almost always prefer a solution that required less maintenance.

tnc1997:
> I don't really want to "pollute" FlutterBluePlus with these changes until we know what we are doing.

Given this and the fact that some of the requests for cross-platform implementations have been open for a couple of years do you think that supporting additional platforms is achievable for the `flutter_blue_plus` plugin?

tnc1997:
> certainly

I am more than happy to support with this but it would be good to decide on the desired platform interface.

chipweinberger:
> I am more than happy to support with this but it would be good to decide on the desired platform interface.

Let's go with the dart layer. highest level.

I think we should still create `flutter_blue_universal` though, with the goal of 
1. making a backwards compatible package to supersede FBP
2. merging it into FBP when it's stable and ready, with the API really figured out

chipweinberger:
Also, I should ask, do you care about Windows? Do you plan to implement it?

tnc1997:
> But I kind of like the idea of a new github repo during development.

So basically using my fork as a development platform interface based implementation in the short-term?

chipweinberger:
Also, I want to reiterate, I'm launching a product next month! :) 

I do want FBP to support more platforms.

But FBP cannot take much of my time right now. Doing this is another repo means I can focus on launching my company.

tnc1997:
> Do you plan to implement it?

I don't have plans currently to implement the Windows platform but it looks like [others](https://github.com/chipweinberger/flutter_blue_plus/pull/858) have attempted this previously.

tnc1997:
> But I do not want to spend time on other platforms personally.

Publishing a platform interface that enables non-endorsed third-party packages for other platforms means that you wouldn't need to. It would be up to said third-parties to maintain the implementations for those other platforms.

chipweinberger:
> Publishing a platform interface that enables non-endorsed third-party packages for other platforms means that you wouldn't need to. It would be up to said third-parties to maintain the implementations for those other platforms.

Assuming we have a good platform interface that supports everyones needs 😉

tnc1997:
> Also, I want to reiterate, I'm launching a product next month!

Wishing you all the very best with the product launch, I hope that it goes well for you and your company :tada:

> When web, or windows, or android adds new methods, it would be ideal if the platform interface does not need to change.

> minimal breakage

> Assuming we have a good platform interface that supports everyones needs 😉

Would you be accepting of a breaking change to existing usages of platform-specific functionality to allow this?

I can work on a proposal for the platform interface that aligns with this if this is something that you're happy with.

chipweinberger:
> Would you be accepting of a breaking change to existing usages of platform-specific functionality to allow this?

You mean like for `FlutterBluePlus.android.requestConnectionPriority(....)`. I think we could just deprecate the current API for this function. Not break it.

chipweinberger:
I think you should look here:

https://pub.dev/packages/bluetooth_low_energy

It gives a good overview for the needs of every platform.

tnc1997:
> You mean like for `FlutterBluePlus.android.requestConnectionPriority(....)`. I think we could just deprecate the current API for this function. No need to break it.

Yeah, basically should the `requestConnectionPriority` method exist in the platform interface and throw an `UnsupportedError` on platforms that do not support it, or should the method only exist in the Android implementation of `flutter_blue_plus` and require using the Android plugin implementation specifically to call the method?

chipweinberger:
> Yeah, basically should the requestConnectionPriority method exist in the platform interface and throw an UnsupportedError on platforms that do not support it, or should the method only exist in the Android implementation of flutter_blue_plus and require using the Android plugin implementation specifically to call the method?

If a minor break means we can avoid a lot of code, I say break it. But if its just a few lines of code to support it, then don't break it.

chipweinberger:
> So basically using my fork as a development platform interface based implementation in the short-term?

Yes. I'd love for a handful of people to try it out first too.

tnc1997:
> And to be clear, I do want to merge it.

That is good to hear because it will require ongoing maintenance overhead on my part to keep the fork in sync.

tnc1997:
> You can see how I've handled deprecations in FBP already. They're all just a few lines of code.

It may not work quite like that code sample because the methods are unsupported and so wouldn't have alternatives.

> But if its just a few lines of code to support it, then don't break it.

In theory it would be as simple as throwing an `UnsupportedError` on those platforms that don't support the method.

Is this something that you would prefer instead of excluding single platform existing functionality from the interface?

e.g. should `requestConnectionPriority` exist in the platform interface package and throw on unsupported platforms.

tnc1997:
> e.g. should `requestConnectionPriority` exist in the platform interface package and throw on unsupported platforms.

I am personally leaning towards including it in the interface and throwing on unsupported platforms.

Having reviewed https://pub.dev/packages/bluetooth_low_energy this is also the approach that is taken there.

chipweinberger:
I say lets just follow the approach of https://github.com/yanshouwang/bluetooth_low_energy

chipweinberger:
When the federated code is ready, I'll merge it, and make you a collaborator here - then we can maintain our respective pieces.

I've put a lot of hours into the code already. Hundreds. So please be mindful of that when making large changes to file structure, etc.

tnc1997:
> I say lets just follow the approach of https://github.com/yanshouwang/bluetooth_low_energy

I will aim to to create a draft platform interface for your initial feedback over the next few days if that's okay.

tnc1997:
> I've put a lot of hours into the code already. Hundreds. So please be mindful of that when making large changes to file structure, etc.

Of course, that's entirely reasonable, I am more than happy for you to lead any refactoring that may be necessary.

tnc1997:
I would like to propose this platform interface that is more aligned with the public static methods in `flutter_blue_plus` and is decoupled from the method channel. When you get a bit of time spare I would welcome your feedback.

chipweinberger:
seems reasonable.

```
  /// Returns a stream of characteristic receive events.
  Stream<OnCharacteristicReceivedEvent> get characteristicReceives {
    throw UnimplementedError();
  }

  /// Returns a stream of characteristic write events.
  Stream<OnCharacteristicWrittenEvent> get characteristicWrites {
    throw UnimplementedError();
  }
```

I'd probably call them `onCharacteristicReceived`, `onCharacteristicWritten`. etc. Makes more sense to me.

chipweinberger:
```
  Stream<OnServicesResetEvent> get servicesChanges {
  Stream<OnDiscoveredServicesEvent> get servicesDiscovers {
```

yes I really dont like these in particular. `onServicesReset`, `onDiscoveredServices` is much better imo.

chipweinberger:
```
  Future<List<int>> readCharacteristic(
    DeviceIdentifier remoteId,
    Guid serviceUuid,
    Guid characteristicUuid,
  ) {
    throw UnimplementedError();
  }
```

must support secondaryServiceUuid as well.

tnc1997:
> I'd call them `onCharacteristicReceived`, `onCharacteristicWritten`. etc. And "on" help signals that it is a stream.

> Yes I really don't like these in particular. `onServicesReset`, `onDiscoveredServices` is much better imo.

No problem, I can rename them accordingly, I was attempting to name them generically not Android specific :smile:

tnc1997:
> must support optional secondaryServiceUuid as well.

Of course, I can add `secondaryServiceUuid`, but I noticed that it is `null` in most usages, is this intentional?

https://github.com/chipweinberger/flutter_blue_plus/blob/dea3493f405820aa400dcfbe48a2d2114e65cea9/lib/src/bluetooth_characteristic.dart#L115
https://github.com/chipweinberger/flutter_blue_plus/blob/dea3493f405820aa400dcfbe48a2d2114e65cea9/lib/src/bluetooth_characteristic.dart#L186
https://github.com/chipweinberger/flutter_blue_plus/blob/dea3493f405820aa400dcfbe48a2d2114e65cea9/lib/src/bluetooth_characteristic.dart#L249

chipweinberger:
theres an outstanding github issue to fix secondary services: https://github.com/chipweinberger/flutter_blue_plus/issues/948

tnc1997:
Thank you very much for your feedback, I have applied your suggestions from the code review.

If you are happy to proceed then I will create auxiliary classes in the platform interface that I will gladly maintain :smile:

chipweinberger:
looks good! just saw this comment

chipweinberger:
Also I wonder if we should have more future proofing

```
  Future<void> callAndroidSpecificFunction({
    String functionName, Map<String, Dynamic>
  }) {
    throw UnimplementedError();
  }
```

```
  Future<void> callAppleSpecificFunction({
    String functionName, Map<String, Dynamic>
  }) {
    throw UnimplementedError();
  }
```

```
  Future<void> callWindowsSpecificFunction({
    String functionName, Map<String, Dynamic>
  }) {
    throw UnimplementedError();
  }
```

chipweinberger:
but im still not that knowledgable about designing these interfaces. I think you know more than me about this.

tnc1997:
> I just added a new option today (https://github.com/chipweinberger/flutter_blue_plus/commit/12c17a6f49c3d6223f7dffaa75f845ccc01c8701) needs to be future proof.

In this scenario we should add that parameter to the method and give it a reasonable default for compatibility.

```dart
Future<void> setOptions({
  bool darwinShowPowerAlert = true,
  bool iosRestoreState = false, // or true
}) {
  throw UnimplementedError();
}
```

> Also I wonder if we should have more future proofing.

Personally I would strongly avoid that because platform interfaces conventionally define properties and/or methods for the functionality. When functionality is added to `flutter_blue_plus`, then we should add an associated method to the platform interface, which should then be implemented by the platform specific implementations that support that functionality. Granted this is a slightly different way of thinking to the method channel which is used as an "interface".

chipweinberger:
let’s say the Windows 12 comes out and adds a few new ble functions. will the platform interface have to change?

or let’s say we want to add more arguments to setOptions. what will the platform implantations all break?

tnc1997:
> let’s say the Windows 12 comes out and adds a few new ble functions. will the platform interface have to change?

In this scenario we should add those new method(s) to the platform interface as a non-breaking change e.g.

```dart
Future<void> newMethod() {
  throw UnimplementedError();
}
```

> or let’s say we want to add more arguments to setOptions. will the platform implantations all break?

If we make a breaking change to the platform interface then this should be published as a breaking change.

If we expect options (for example) to change often then we should probably consider something like this:

```dart
class Options {
  final bool darwinShowPowerAlert;
  final bool additionalParameter;

  Options({
    this.darwinShowPowerAlert = true,
    this.additionalParameter = false, // or true
  });
}

abstract base class FlutterBluePlusPlatform {
  Future<void> setOptions(
    Options options,
  ) {
    throw UnimplementedError();
  }
}
```

chipweinberger:
let's make the platform interface as future proof as possible 

options will evolve over time 

i don't want to have to maintain windows, web, linux

so the platform interface should allow them enough flexibility to add new functions without my involvement

tnc1997:
> let's make the platform interface as future proof as possible

> so the platform interface should allow them enough flexibility to add new functions without my involvement

Just to confirm that the platform interface may have methods added to it in the future to cover potential functionality.

One of the reasons for a platform interface is that it is the interface used by all platforms not just Android/iOS/macOS.

tnc1997:
I would be more than happy to maintain the platform interface in collaboration with you to reduce your burden.

Just double checking that you're happy to continue with us creating a platform interface for `flutter_blue_plus`?

chipweinberger:
what is the status of being able to do things like this?

`FlutterBluePlus.android.clearGattCache()`? or `FlutterBluePlus.android.someOtherMethod()`

tnc1997:
> what is the status of being able to do things like this?

Researching other examples of platform interfaces that typically isn't how they are created.

* https://github.com/yanshouwang/bluetooth_low_energy/blob/master/bluetooth_low_energy_platform_interface/lib/src/bluetooth_low_energy_manager.dart
* https://github.com/PhilipsHue/flutter_reactive_ble/blob/master/packages/reactive_ble_platform_interface/lib/src/reactive_ble_platform_interface.dart

> FlutterBluePlus.android.clearGattCache()

This implies that each platform has its own interface and so isn't a shared platform interface.

> Platform implementations should extends their platform interface class rather than implements it, as newly added methods to platform interfaces are not considered breaking changes. Extending a platform interface ensures that subclasses will get the default implementations from the base class, while platform implementations that implements their platform interface will be broken by newly added methods.

tnc1997:
> FlutterBluePlus.android.clearGattCache()

That isn't to say that you couldn't do that in the existing `flutter_blue_plus` package, but looking at other examples of platform interfaces you wouldn't do that in a `flutter_blue_plus_platform_interface` package. For example you could have a method like that in `flutter_blue_plus` that calls `FlutterBluePlusPlatform.instance.clearGattCache()`.

chipweinberger:
what would the user facing API be? If not `FlutterBluePlus.windows.someNewMethod()`?

Maybe `FlutterBluePlusWindows.someNewMethod()`?

tnc1997:
> what would the user facing API be? If not `FlutterBluePlus.windows.someNewMethod()`?
> 
> Maybe `FlutterBluePlusWindows.someNewMethod()`?

The methods in the platform interface could be as per the changes in this pull request, whilst the methods in `flutter_blue_plus` would be as they are now, albeit with implementations that use the platform interface instead of a method channel.

The methods in the platform interface should then be implemented by each platform specific implementation where appropriate and possible, throwing an unsupported error if that particular method is not supported on the platform.

chipweinberger:
Maybe im not clear. Does the platform interface *need* to be a superset of all platforms?

If so, that's pretty annoying because every addition will require the platform interface to change.

chipweinberger:
Have you seen this package? https://pub.dev/packages/flutter_blue_plus_windows

He added Windows support to FBP without any involvement from me. 

He used a wrapper architecture:
https://github.com/chan150/flutter_blue_plus_windows/blob/master/lib/src/wrapper/flutter_blue_plus_wrapper.dart

Perhaps you can add linux & web in a similar fashion?

tnc1997:
> Maybe im not clear. Does the platform interface need to be a superset of all platforms?

Yes, the platform interface should be a superset of all platforms, otherwise it is not a "true" platform interface. I would encourage you to take a look at the platform interfaces that I previously linked to and those [here](https://github.com/flutter/packages/tree/main) for inspiration.

When adding a method to one or more platforms, it should be added to the platform interface, then implemented in the platforms that support that particular functionality.

chipweinberger:
> Yes, the platform interface should be a superset of all platforms, otherwise it is not a "true" platform interface.

Unfortunately, this is just not something I'm comfortable maintaining long term. I know I wont have enough enthusiasm make each platform a first class citizen in the platform interface.

> I would be more than happy to maintain the platform interface in collaboration with you to reduce your burden.

I appreciate that offer! But I think then FBP for android/ios/mac would not be fully autonomous? That would make development of the main platforms harder.

----

I think you should consider that approach taken by https://pub.dev/packages/flutter_blue_plus_windows. It's much more autonomous, and the "wrapper" is not that difficult to write/maintain.

tnc1997:
> I think you should consider that approach taken by https://pub.dev/packages/flutter_blue_plus_windows. It's much more autonomous, and the \"wrapper\" is not that difficult to write/maintain.

Unfortunately as Stuart highlighted in [this](https://github.com/flutter/flutter/issues/143095) issue, that approach has a number of significant disadvantages including being very fragile and prone to breakage. This makes ongoing maintenance of the platform implementation a lot more complicated compared to simply implementing a platform interface. It is a shame that `flutter_blue_plus` will not create and publish a platform interface as per other popular plugins at this time.

chipweinberger:
> It is a shame that flutter_blue_plus will not create and publish a platform interface as per other popular plugins at this time.

I share your disappointment. But I just don't have the bandwidth to do it justice. 

I welcome you to create `flutter_blue_plus_universal` though, if you have the energy. You can even wrap FBP, so that you don't need to maintain android/mac/iOS yourself. Then you can just "peg" your `pubspec.yaml` to depend on a _specific_ FBP version, and it will never break. FBP is *rather* stable at this point.

Issue #954, opened 2024-08-05
[Feature]: read/write/notify value as raw bytes


boborua:
### FlutterBluePlus Version

1.12.7

### Flutter Version

3.13.9

### What OS?

Android, iOS

### OS Version

iOS 17 / Android 33

### Bluetooth Module

*

### What is your feature request?

It is limited from the type of api arguments when we use blue_pluse reading/writing value, for example write(List<int> value) on a Charateristic.

After having done some research in the repository, I found that in Android platform, data was received in byte[] format (Java side) and encoded to hex string to pass to the dart side.

Is it possible that we can write and read value in Uint8Array/List<Uint8>...etc 

### Logs

```shell
no logs
```

chipweinberger:
not possible.

PR #1010, opened 2024-10-06
feat: uint8list instead of int list

evan361425:
## Purpose

Using [`Uint8List`](https://api.flutter.dev/flutter/dart-typed_data/Uint8List-class.html) to pass the value as proposed in #954.

## Why?

To avoid converting between strings and byte lists. There are also several benefits shown in this [video](https://www.youtube.com/watch?v=9lhN5QXyZQc).

## Tests

I've tested this on Android, and it works as expected, but I haven't tested it on iOS (I don't have a physical device).

## Breaking changes

We can easily convert between `Uint8List` and `List<int>`:

```dart
final v = <int>[1, 2, 3];
print(v.join(',')); // 1,2,3
final vv = Uint8List.fromList(v);
print(vv.join(',')); // 1,2,3
final vvv = vv as List<int>;
print(vvv.join(',')); // 1,2,3
```

However, this is still a breaking change. How can we make the migration painless? We could provide a value8 in the property and deprecated getter of value8, for example:

```dart
class BmDescriptorData {
  final Uint8List value8;

  factory BmDescriptorData.fromMap(Map<dynamic, dynamic> json) {
    return BmDescriptorData(
      value8: json['value'],
      // ...
    );
  }

  // @deprecate
  List<int> get value => value8 as List<int>;
}
```

But I'm not sure whether we should offer this migration path or simply bump the major version.

### DIFF
--- a/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -2279,7 +2279,7 @@ public class FlutterBluePlusPlugin implements
             LogLevel level = LogLevel.DEBUG;
             log(level, "onCharacteristicChanged:");
             log(level, "  chr: " + uuidStr(characteristic.getUuid()));
-            onCharacteristicReceived(gatt, characteristic, value, status);
+            onCharacteristicReceived(gatt, characteristic, value, BluetoothGatt.GATT_SUCCESS);
         }
 
         @Override
@@ -2317,7 +2317,7 @@ public class FlutterBluePlusPlugin implements
 
             // what data did we write?
             String key = remoteId + ":" + serviceUuid + ":" + characteristicUuid;
-            []byte value = mWriteChr.get(key) != null ? mWriteChr.get(key) : new byte[0];
+            byte[] value = mWriteChr.get(key) != null ? mWriteChr.get(key) : new byte[0];
             mWriteChr.remove(key);
 
             // see: BmCharacteristicData
--- a/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -106,8 +106,8 @@ public class FlutterBluePlusPlugin implements
     private final Map<String, BluetoothDevice> mBondingDevices = new ConcurrentHashMap<>();
     private final Map<String, Integer> mMtu = new ConcurrentHashMap<>();
     private final Map<String, BluetoothGatt> mAutoConnected = new ConcurrentHashMap<>();
-    private final Map<String, String> mWriteChr = new ConcurrentHashMap<>();
-    private final Map<String, String> mWriteDesc = new ConcurrentHashMap<>();
+    private final Map<String, byte[]> mWriteChr = new ConcurrentHashMap<>();
+    private final Map<String, byte[]> mWriteDesc = new ConcurrentHashMap<>();
     private final Map<String, String> mAdvSeen = new ConcurrentHashMap<>();
     private final Map<String, Integer> mScanCounts = new ConcurrentHashMap<>();
     private HashMap<String, Object> mScanFilters = new HashMap<String, Object>();
@@ -370,7 +370,7 @@ public class FlutterBluePlusPlugin implements
                     break;
                 }
 
-               case "getAdapterName":
+                case "getAdapterName":
                 {
                     ArrayList<String> permissions = new ArrayList<>();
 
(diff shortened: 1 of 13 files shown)


android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -2279,7 +2279,7 @@ public void onCharacteristicChanged(BluetoothGatt gatt, BluetoothGattCharacteris
             LogLevel level = LogLevel.DEBUG;
             log(level, "onCharacteristicChanged:");
             log(level, "  chr: " + uuidStr(characteristic.getUuid()));
-            onCharacteristicReceived(gatt, characteristic, value, status);
+            onCharacteristicReceived(gatt, characteristic, value, BluetoothGatt.GATT_SUCCESS);
```
evan361425, line 2282: This is a bug, using status which is undefined.

chipweinberger:
unfortunately, these are breaking changes to the API.

so I wont merge.

its a good idea though. I didnt design the original API...

evan361425:
Will it be considered if I make it backward compatible?

chipweinberger:
you would need to measure perf impact at the minimum

evan361425:
OK, I'll try it but there's still a [bug](https://github.com/chipweinberger/flutter_blue_plus/pull/1010#discussion_r1788858263) anyway.

android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
chipweinberger: fixed

chipweinberger:
hi, did you measure the performance impact yet?

evan361425:
Still in plan😂

evan361425:
@chipweinberger Thanks for the patience, I've test it in [performance_screen](https://github.com/evan361425/flutter_blue_plus/blob/perf/atoi/example/lib/screens/performance_screen.dart), with below data:

- Request size: 512 bytes
- Each epoch sending requests count: 100
- Epoch count: 50

And here is the result in seconds:

| Type | Avg | P90 | P95 | P99 |
| - | - | - | - | - |
| List of int to string | 0.84634 | 0.9288 | 1.0356 | 1.1860 |
| Uint8List | 0.82896 | 0.9449 | 0.9887 | 1.1166 |

<details>
<summary>Raw data</summary>

List of int to string:

```text
1.1610
0.99600
0.92600
0.91000
0.86700
0.89700
0.88200
0.88200
0.87800
0.91200
0.89900
0.80400
0.83000
0.80000
0.86400
0.86200
0.81500
0.83200
0.81700
0.83700
0.83300
0.81000
0.85800
0.95400
1.0680
0.80700
1.2100
0.87200
0.79000
0.82100
0.76800
0.80500
0.76900
0.77800
0.79800
0.74600
0.74300
0.74200
0.76300
0.78600
0.79600
0.77700
0.75900
0.89600
0.77900
0.74600
0.74700
0.85000
0.81300
0.76200
```

Uint8List:

```text
1.0630
0.96200
0.88600
0.90900
1.1680
0.98100
0.94300
0.99500
0.82500
0.80800
0.82900
0.81400
0.91300
0.88500
0.93700
0.82300
0.80400
0.77900
0.79100
0.76500
0.77600
0.78700
0.79200
0.81400
0.81300
0.80200
0.81200
0.77700
0.78800
0.81800
0.74700
0.79800
0.78700
0.78700
0.78500
0.77100
0.80000
0.77800
0.75800
0.80200
0.71300
0.81300
0.76600
0.77900
0.84500
0.73600
0.81300
0.75800
0.80700
0.74600
```
</details>

chipweinberger:
thanks for the numbers!!

a nice little improvement! but not enough of a difference to break api.

chipweinberger:
> but we could have a small queue on the native side, to avoid round tripping through Dart every time.

this is an idea i've wanted to try for awhile. it would be annoying to implement. but easy enough to test to see how much it improves things.

test #1: do lots of writes with FBP
test #2: do lots of writes using native only 

and compare.

chipweinberger:
> we could avoid converting to strings in the internal code, and probably get similar improvements.
> but i like hex, it's more useful for during application development, logs, debugging issues people file on github, etc.

thinking about this more, we could still log hex for verbose logs. we just need to update the logging code on the dart side to convert Uint8List to hex strings before we log.

```
      // log args
      if (logLevel == LogLevel.verbose) {
        String func = '<$method>';
        String args = arguments.toString();
        func = _logColor ? _black(func) : func;
        args = _logColor ? _magenta(args) : args;
        print("[FBP] $func args: $args");
      }

      // invoke
      out = await _methodChannel.invokeMethod(method, arguments);

      // log result
      if (logLevel == LogLevel.verbose) {
        String func = '($method)';
        String result = out.toString();
        func = _logColor ? _black(func) : func;
        result = _logColor ? _brown(result) : result;
        print("[FBP] $func result: $result");
      }
```

evan361425:
I'll try, thanks.

evan361425:
BTW, the result above may be the baseline in future improvements? The performance screen can do e2e test and may have a better performance result.

PR #1042, opened 2024-11-06
[Feature] Support multiple service/characteristic instances with same UUID

### LINKED ISSUE #795: [Feature]: Support multiple characteristics with same uuid
### FlutterBluePlus Version

1.31.14

### Flutter Version

3.18.1

### What OS?

All

### OS Version

All

### Bluetooth Module

All

### What is your feature request?

Relevant Issue: https://github.com/boskokg/flutter_blue_plus/issues/789

**FlutterBluePlus does not yet support multiple characteristics with the same uuid**, which is needed for the 0x1812 HID Service.

We need to add a new class member to `BluetoothCharacteristic`: `final int characteristicIndex`

For most characteristics, index will be 0. But If there are multiple characteristics with the same uuid, it represents if it is the first, second, third, etc.  Android & iOS always keep characteristics in the same order, so we can use an index to find the right one.

---

> in the BLE 0x1812 service is it possible to have multiple 0x2a4d characteristics?
> 
> Yes, in the Bluetooth Low Energy (BLE) specification, the Human Interface Device (HID) service, identified by the UUID 0x1812, can indeed have multiple instances of the Report characteristic, which is identified by the UUID 0x2A4D. This design allows a single HID service to support multiple reports, such as keyboard input, mouse input, and other control or data reports, each of which can be defined by its own instance of the Report characteristic. 
> 
> This capability is crucial for devices that serve multiple functions or support complex interactions, allowing them to report different types of data to a host device under a single HID service. Each 0x2A4D characteristic instance can be configured with different properties (like Read, Write, Notify) and permissions, and it can be associated with different Report Reference descriptors that define the type and ID of the report, enabling the host to distinguish between them.

### Logs

```shell
No Logs
```

nebkat:
Untested on iOS, working on Android.

Based on https://github.com/pauldemarco/flutter_blue/pull/215

Closes #795

### DIFF
--- a/lib/src/bluetooth_attribute.dart
+++ b/lib/src/bluetooth_attribute.dart
@@ -1,0 +1,77 @@
+// Copyright 2017-2023, Charles Weinberger & Paul DeMarco.
+// All rights reserved. Use of this source code is governed by a
+// BSD-style license that can be found in the LICENSE file.
+
+part of flutter_blue_plus;
+
+abstract class BluetoothAttribute {
+  final BluetoothDevice device;
+  final Guid uuid;
+  final int? index;
+
+  BluetoothAttribute({
+    required this.device,
+    required this.uuid,
+    this.index,
+  });
+
+  DeviceIdentifier get remoteId => device.remoteId;
+
+  @Deprecated('Use remoteId instead')
+  DeviceIdentifier get deviceId => remoteId;
+
+  BluetoothAttribute? get _parentAttribute => null;
+
+  String get identifier => "$uuid:$index";
+
+  String get identifierPath =>
+      _parentAttribute != null ? "${_parentAttribute!.identifierPath}/$identifier" : identifier;
+}
+
+abstract class BluetoothValueAttribute extends BluetoothAttribute {
+  List<int> _lastValue = [];
+
+  BluetoothValueAttribute({
+    required BluetoothDevice device,
+    required Guid uuid,
+    int? index,
+  }) : super(device: device, uuid: uuid, index: index);
+
+  /// this variable is updated:
+  ///   - anytime `read()` is called
+  ///   - anytime `write()` is called
+  ///   - anytime a notification arrives (characteristics, if subscribed)
+  ///   - when the device is disconnected it is cleared
+  List<int> get lastValue => _lastValue;
+
+  /// this stream emits values:
+  ///   - anytime `read()` is called
+  ///   - anytime `write()` is called
+  ///   - anytime a notification arrives (characteristics, if subscribed)
+  ///   - and when first listened to, it re-emits the last value for convenience
+  Stream<List<int>> get lastValueStream => FlutterBluePlus._methodStream.stream
+      .where((e) =>
+          e is OnCharacteristicReceivedEvent ||
+          e is OnCharacteristicWrittenEvent ||
+          e is OnDescriptorReadEvent ||
+          e is OnDescriptorWrittenEvent)
+      .map((e) => e as GetAttributeValueMixin)
+      .where((e) => e.attribute == this)
+      .map((e) => e.value)
+      .newStreamWithInitialValue(lastValue);
+
+  /// this stream emits values:
+  ///   - anytime `read()` is called
+  ///   - anytime a notification arrives (if subscribed)
+  Stream<List<int>> get onValueReceived => FlutterBluePlus._methodStream.stream
+      .where((e) => e is OnCharacteristicReceivedEvent || e is OnDescriptorReadEvent)
+      .map((e) => e as GetAttributeValueMixin)
+      .where((e) => e.attribute == this)
+      .map((e) => e.value);
+
+  @Deprecated('Use onValueReceived instead')
+  Stream<List<int>> get value => onValueReceived;
+
+  @Deprecated('Use onValueReceived instead')
+  Stream<List<int>> get onValueChangedStream => onValueReceived;
+}
--- a/lib/src/flutter_blue_plus.dart
+++ b/lib/src/flutter_blue_plus.dart
@@ -16,20 +16,10 @@ class FlutterBluePlus {
 
   /// a broadcast stream version of the MethodChannel
   // ignore: close_sinks
-  static final StreamController<MethodCall> _methodStream = StreamController.broadcast();
+  static final StreamController<dynamic> _methodStream = StreamController.broadcast();
 
   // always keep track of these device variables
-  static final Map<DeviceIdentifier, BmConnectionStateResponse> _connectionStates = {};
-  static final Map<DeviceIdentifier, BmDiscoverServicesResult> _knownServices = {};
-  static final Map<DeviceIdentifier, BmBondStateResponse> _bondStates = {};
-  static final Map<DeviceIdentifier, BmMtuChangedResponse> _mtuValues = {};
-  static final Map<DeviceIdentifier, String> _platformNames = {};
-  static final Map<DeviceIdentifier, String> _advNames = {};
-  static final Map<DeviceIdentifier, Map<String, List<int>>> _lastChrs = {};
-  static final Map<DeviceIdentifier, Map<String, List<int>>> _lastDescs = {};
-  static final Map<DeviceIdentifier, List<StreamSubscription>> _deviceSubscriptions = {};
-  static final Map<DeviceIdentifier, List<StreamSubscription>> _delayedSubscriptions = {};
-  static final Map<DeviceIdentifier, DateTime> _connectTimestamp = {};
+  static final Map<DeviceIdentifier, BluetoothDevice> _devices = {};
   static final List<StreamSubscription> _scanSubscriptions = [];
   static final Set<DeviceIdentifier> _autoConnect = {};
 
@@ -40,16 +30,16 @@ class FlutterBluePlus {
   static final _scanResults = _StreamControllerReEmit<List<ScanResult>>(initialValue: []);
 
   /// buffers the scan results
-  static _BufferStream<BmScanResponse>? _scanBuffer;
+  static _BufferStream<OnScanResponseEvent>? _scanBuffer;
 
   /// the subscription to the merged scan results stream
-  static StreamSubscription<BmScanResponse?>? _scanSubscription;
+  static StreamSubscription<OnScanResponseEvent?>? _scanSubscription;
 
   /// timeout for scanning that can be cancelled by stopScan
   static Timer? _scanTimeout;
 
   /// the last known adapter state
-  static BmAdapterStateEnum? _adapterStateNow;
+  static BluetoothAdapterState? _adapterStateNow;
 
   /// FlutterBluePlus log level
   static LogLevel _logLevel = LogLevel.debug;
@@ -65,8 +55,7 @@ class FlutterBluePlus {
   static Future<bool> get isSupported async => await _invokeMethod('isSupported');
 
   /// The current adapter state
-  static BluetoothAdapterState get adapterStateNow =>
-      _adapterStateNow != null ? _bmToAdapterState(_adapterStateNow!) : BluetoothAdapterState.unknown;
+  static BluetoothAdapterState get adapterStateNow => _adapterStateNow ?? BluetoothAdapterState.unknown;
 
   /// Return the friendly Bluetooth name of the local Bluetooth adapter
   static Future<String> get adapterName async => await _invokeMethod('getAdapterName');
@@ -117,13 +106,10 @@ class FlutterBluePlus {
 
   /// Turn on Bluetooth (Android only),
   static Future<void> turnOn({int timeout = 60}) async {
-    var responseStream = FlutterBluePlus._methodStream.stream
-        .where((m) => m.method == "OnTurnOnResponse")
-        .map((m) => m.arguments)
-        .map((args) => BmTurnOnResponse.fromMap(args));
+    var responseStream = FlutterBluePlus._extractEventStream<OnTurnOnResponseEvent>();
 
     // Start listening now, before invokeMethod, to ensure we don't miss the response
-    Future<BmTurnOnResponse> futureResponse = responseStream.first;
+    Future<OnTurnOnResponseEvent> futureResponse = responseStream.first;
 
     // invoke
     bool changed = await _invokeMethod('turnOn');
@@ -131,7 +117,7 @@ class FlutterBluePlus {
     // only wait if bluetooth was off
     if (changed) {
       // wait for response
-      BmTurnOnResponse response = await futureResponse.fbpTimeout(timeout, "turnOn");
+      OnTurnOnResponseEvent response = await futureResponse.fbpTimeout(timeout, "turnOn");
 
       // check response
       if (response.userAccepted == false) {
@@ -151,24 +137,17 @@ class FlutterBluePlus {
       var value = BmBluetoothAdapterState.fromMap(result).adapterState;
       // update _adapterStateNow if it is still null after the await
       if (_adapterStateNow == null) {
-        _adapterStateNow = value;
+        _adapterStateNow = _bmToAdapterState(value);
       }
     }
 
-    yield* FlutterBluePlus._methodStream.stream
-        .where((m) => m.method == "OnAdapterStateChanged")
-        .map((m) => m.arguments)
-        .map((args) => BmBluetoothAdapterState.fromMap(args))
-        .map((s) => _bmToAdapterState(s.adapterState))
-        .newStreamWithInitialValue(_bmToAdapterState(_adapterStateNow!));
+    yield* FlutterBluePlus._extractEventStream<OnAdapterStateChangedEvent>()
+        .map((s) => s.adapterState)
+        .newStreamWithInitialValue(_adapterStateNow!);
   }
 
   /// Retrieve a list of devices currently connected to your app
-  static List<BluetoothDevice> get connectedDevices {
-    var copy = Map.from(_connectionStates);
-    copy.removeWhere((key, value) => value.connectionState == BmConnectionStateEnum.disconnected);
-    return copy.values.map((v) => BluetoothDevice(remoteId: v.remoteId)).toList();
-  }
+  static List<BluetoothDevice> get connectedDevices => _devices.values.where((d) => d.isConnected).toList();
 
   /// Retrieve a list of devices currently connected to the system
   /// - The list includes devices connected to by *any* app
@@ -177,24 +156,14 @@ class FlutterBluePlus {
   static Future<List<BluetoothDevice>> systemDevices(List<Guid> withServices) async {
     var result = await _invokeMethod('getSystemDevices', {"with_services": withServices.map((s) => s.str).toList()});
     var r = BmDevicesList.fromMap(result);
-    for (BmBluetoothDevice device in r.devices) {
-      if (device.platformName != null) {
-        _platformNames[device.remoteId] = device.platformName!;
-      }
-    }
-    return r.devices.map((d) => BluetoothDevice.fromId(d.remoteId.str)).toList();
+    return r.devices.map((d) => FlutterBluePlus._deviceForId(d.remoteId).._platformName = d.platformName).toList();
   }
 
   /// Retrieve a list of bonded devices (Android only)
   static Future<List<BluetoothDevice>> get bondedDevices async {
     var result = await _invokeMethod('getBondedDevices');
     var r = BmDevicesList.fromMap(result);
-    for (BmBluetoothDevice device in r.devices) {
-      if (device.platformName != null) {
-        _platformNames[device.remoteId] = device.platformName!;
-      }
-    }
-    return r.devices.map((d) => BluetoothDevice.fromId(d.remoteId.str)).toList();
+    return r.devices.map((d) => FlutterBluePlus._deviceForId(d.remoteId).._platformName = d.platformName).toList();
   }
 
   /// Start a scan, and return a stream of results
@@ -282,10 +251,7 @@ class FlutterBluePlus {
           androidScanMode: androidScanMode.value,
           androidUsesFineLocation: androidUsesFineLocation);
 
-      Stream<BmScanResponse> responseStream = FlutterBluePlus._methodStream.stream
-          .where((m) => m.method == "OnScanResponse")
-          .map((m) => m.arguments)
-          .map((args) => BmScanResponse.fromMap(args));
+      Stream<OnScanResponseEvent> responseStream = FlutterBluePlus._extractEventStream<OnScanResponseEvent>();
 
       // Start listening now, before invokeMethod, so we do not miss any results
       _scanBuffer = _BufferStream.listen(responseStream);
@@ -297,7 +263,7 @@ class FlutterBluePlus {
       });
 
       // check every 250ms for gone devices?
-      late Stream<BmScanResponse?> outputStream = removeIfGone != null
+      late Stream<OnScanResponseEvent?> outputStream = removeIfGone != null
           ? _mergeStreams([_scanBuffer!.stream, Stream.periodic(Duration(milliseconds: 250))])
           : _scanBuffer!.stream;
 
@@ -307,7 +273,7 @@ class FlutterBluePlus {
       List<ScanResult> output = [];
 
       // listen & push to `scanResults` stream
-      _scanSubscription = outputStream.listen((BmScanResponse? response) {
+      _scanSubscription = outputStream.listen((OnScanResponseEvent? response) {
         if (response == null) {
           // if null, this is just a periodic update to remove old results
           if (output._removeWhere((elm) => DateTime.now().difference(elm.timeStamp) > removeIfGone!)) {
@@ -315,26 +281,23 @@ class FlutterBluePlus {
           }
         } else {
           // failure?
-          if (response.success == false) {
-            var e = FlutterBluePlusException(_nativeError, "scan", response.errorCode, response.errorString);
-            _scanResults.addError(e);
+          final exception = response.exception("scan");
+          if (exception != null) {
+            _scanResults.addError(exception);
             _stopScan(invokePlatform: false);
           }
 
           // iterate through advertisements
-          for (BmScanAdvertisement bm in response.advertisements) {
+          for (ScanResult sr in response.advertisements) {
             // cache platform name
-            if (bm.platformName != null) {
-              _platformNames[bm.remoteId] = bm.platformName!;
-            }
+            // TODO CRITICAL if (sr.advertisementData.platformName) {
+            //   _platformNames[bm.remoteId] = bm.platformName!;
+            // }
 
             // cache advertised name
-            if (bm.advName != null) {
-              _advNames[bm.remoteId] = bm.advName!;
-            }
-
-            // convert
-            ScanResult sr = ScanResult.fromProto(bm);
+            // TODO CRITICAL if (bm.advName != null) {
+            //   _advNames[bm.remoteId] = bm.advName!;
+            // }
 
             if (oneByOne) {
               // push single item
@@ -417,6 +380,10 @@ class FlutterBluePlus {
     return await _invokeMethod('getPhySupport').then((args) => PhySupport.fromMap(args));
   }
 
+  static BluetoothDevice _deviceForId(DeviceIdentifier id) {
+    return _devices.putIfAbsent(id, () => BluetoothDevice._internal(remoteId: id));
+  }
+
   static Future<dynamic> _initFlutterBluePlus() async {
     if (_initialized) {
       return;
@@ -425,7 +392,14 @@ class FlutterBluePlus {
     _initialized = true;
 
     // set platform method handler
-    _methodChannel.setMethodCallHandler(_methodCallHandler);
+    _methodChannel.setMethodCallHandler((call) async {
+      try {
+        return await _methodCallHandler(call);
+      } catch (e, s) {
+        print("[FBP] Error in methodCallHandler: $e $s");
+        rethrow;
+      }
+    });
 
     // flutter restart - wait for all devices to disconnect
     if ((await _methodChannel.invokeMethod('flutterRestart')) != 0) {
@@ -433,6 +407,50 @@ class FlutterBluePlus {
       while ((await _methodChannel.invokeMethod('connectedCount')) != 0) {
         await Future.delayed(Duration(milliseconds: 50));
       }
+    }
+  }
+
+  static dynamic _methodCallMap(MethodCall call) {
+    if (call.method == OnDetachedFromEngineEvent.method) {
+      return OnDetachedFromEngineEvent();
+    } else if (call.method == OnDiscoveredServicesEvent.method) {
+      BmDiscoverServicesResult r = BmDiscoverServicesResult.fromMap(call.arguments);
+      return OnDiscoveredServicesEvent(r);
+    } else if (call.method == OnAdapterStateChangedEvent.method) {
+      BmBluetoothAdapterState r = BmBluetoothAdapterState.fromMap(call.arguments);
+      return OnAdapterStateChangedEvent(r);
+    } else if (call.method == OnConnectionStateChangedEvent.method) {
+      BmConnectionStateResponse r = BmConnectionStateResponse.fromMap(call.arguments);
+      return OnConnectionStateChangedEvent(r);
+    } else if (call.method == OnBondStateChangedEvent.method) {
+      BmBondStateResponse r = BmBondStateResponse.fromMap(call.arguments);
+      return OnBondStateChangedEvent(r);
+    } else if (call.method == OnNameChangedEvent.method) {
+      BmNameChanged r = BmNameChanged.fromMap(call.arguments);
+      return OnNameChangedEvent(r);
+    } else if (call.method == OnServicesResetEvent.method) {
+      BmBluetoothDevice r = BmBluetoothDevice.fromMap(call.arguments);
+      return OnServicesResetEvent(r);
+    } else if (call.method == OnMtuChangedEvent.method) {
+      BmMtuChangedResponse r = BmMtuChangedResponse.fromMap(call.arguments);
+      return OnMtuChangedEvent(r);
+    } else if (call.method == OnCharacteristicReceivedEvent.method) {
+      BmCharacteristicData r = BmCharacteristicData.fromMap(call.arguments);
+      return OnCharacteristicReceivedEvent(r);
+    } else if (call.method == OnCharacteristicWrittenEvent.method) {
+      BmCharacteristicData r = BmCharacteristicData.fromMap(call.arguments);
+      return OnCharacteristicWrittenEvent(r);
+    } else if (call.method == OnDescriptorReadEvent.method) {
+      BmDescriptorData r = BmDescriptorData.fromMap(call.arguments);
+      return OnDescriptorReadEvent(r);
+    } else if (call.method == OnDescriptorWrittenEvent.method) {
+      BmDescriptorData r = BmDescriptorData.fromMap(call.arguments);
+      return OnDescriptorWrittenEvent(r);
+    } else if (call.method == OnScanResponseEvent.method) {
+      BmScanResponse r = BmScanResponse.fromMap(call.arguments);
+      return OnScanResponseEvent(r);
+    } else {
+      throw UnimplementedError("methodCallMap: ${call.method}");
     }
   }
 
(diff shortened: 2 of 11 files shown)


chipweinberger:
wow thanks!! this looks pretty good.

1. please test on iOS. you can buy an old iOS device for about $150. 

2. use shorter named.  `serviceIdx`,`secondaryServiceIdx`, `characteristicIndex`

3.  hashes are not guaranteed to be unique. so your code will not work sometimes. you should just use it's array location, i.e. index. e.g.

```
for (int serviceIdx = 0; serviceIdx < services.length; serviceIdx++);
```

nebkat:
> 1. please test on iOS. you can buy an old iOS device for about $150.

Is it possible to build for iOS on a Windows machine? Anyway I've since had a colleague confirm that this is working as expected on iOS!

> 2. use shorter names.  `serviceIndex`,`secondaryServiceIndex`, `characteristicIndex`

Currently this is named after the Android `instanceId`. I was thinking it could be better to make a `Guid`/`int` pair type, perhaps called `BluetoothAttribute`, and use it everywhere where we have the two? Can also provide getters to still be able to access the `uuid` property directly.

> 3. hashes are not guaranteed to be unique. so your code will not always work. you should just use it's array location, i.e. index. e.g.
> 
> ```
> for (int serviceIndex = 0; serviceIndex < services.length; serviceIndex++);
> ```

Can we be certain that the order will never change? Based on https://github.com/onmyway133/Runtime-Headers/blob/master/macOS/10.12/CoreBluetooth.framework/CBCharacteristic.h I suspect the internal hash function is just combining the handle and the uuid - so unless there are hundreds of characteristics with the same UUID it is highly unlikely they would collide.

> 4. in dart, the index should default to 0
> 
> ```
> class BluetoothCharacteristic {
>   final DeviceIdentifier remoteId;
>   final Guid serviceUuid;
>   final int serviceIndex;
>   final Guid? secondaryServiceUuid;
>   final int? secondaryServiceIndex;
>   final Guid characteristicUuid;
>   final int characteristicIndex;
> 
>   BluetoothCharacteristic({
>     required this.remoteId,
>     required this.serviceUuid,
>     this.serviceIndex = 0,
>     this.secondaryServiceUuid,
>     this.secondaryServiceIndex = 0,
>     required this.characteristicUuid,
>     required this.characteristicIndex = 0,
>   });
> ```

Sure - will do unless you like my proposal for `BluetoothAttribute`.

chipweinberger:
> Is it possible to build for iOS on a Windows machine? 

ah good point. I don't think so.

> Currently this is named after the Android instanceId. I was thinking it could be better to make a Guid/int pair type, perhaps called BluetoothAttribute, and use it everywhere where we have the two? Can also provide getters to still be able to access the uuid property directly.

I think just keep it simple. What you have now is good. 

I just dont like the name. it's longer, more complicated, not available on iOS. But I understand it makes sense if this feature was android-only.

> Can we be certain that the order will never change?

Yes. They guarantee order on iOS, because it's important for things like this.

nebkat:
Have completed the changes and will have colleague verify iOS again.

The `[characteristic.service.characteristics indexOfObject:characteristic]` everywhere feels somewhat convoluted but I believe this will achieve the desired effect.

chipweinberger:
I have been refactoring secondary services, so we are going to have a merge conflict, sorry about that.

I'm about to push the changes to master branch.

chipweinberger:
here is the commit. 

https://github.com/chipweinberger/flutter_blue_plus/commit/2064c18e40d1a01e2bf1e382c746acc588cc029f

please read this comment for context: https://github.com/chipweinberger/flutter_blue_plus/issues/948#issuecomment-2461056547

nebkat:
No worries, I was actually considering proposing that exact change so I am glad to see it! Will refactor tomorrow.

chipweinberger:
also, you'll need to mirror some of the changes I made.

1. always use the same order. I use, in order: remoteId, svcUuid, chrUuid, descUuid, primarySvcUuid

for you, you should do: remoteId, svcUuid, svcIndex, chrUuid, chrIndex, descUuid, primarySvcUuid

2. you need to update code that makes keys

```
                    String key = remoteId + ":" + serviceUuid + ":" + characteristicUuid + ":" + descriptorUuid + ":" + primaryServiceUuid;
                    ```

e..g `mWriteDesc`, `mWriteChr,` and the same for iOS: `writeChrs`, `writeDesc`

**Comment by chipweinberger** [comment]
also make sure you update this code, to check that the indexes are equal

```
  // get known service
  BmBluetoothService? get _bmsvc {
    if (FlutterBluePlus._knownServices[remoteId] != null) {
      for (var s in FlutterBluePlus._knownServices[remoteId]!.services) {
        if (s.serviceUuid == serviceUuid) {
          if (s.primaryServiceUuid == primaryServiceUuid) {
            return s;
          }
        }
      }
    }
    return null;
  }

  /// get known characteristic
  BmBluetoothCharacteristic? get _bmchr {
    if (_bmsvc != null) {
      for (var c in _bmsvc!.characteristics) {
        if (c.characteristicUuid == uuid) {
          return c;
        }
      }
    }
    return null;
  }
  ```

nebkat:
Although having had a quick look at it I'm not sure I expected `includedServices` to be removed. I think that was a good representation of the service hierarchy as it is defined in GATT.

Could we make it bi-directional by retaining `includedServices` alongside `primaryServiceUuid`? It would just involve initializing all `BluetoothService` instances first with `primaryServiceUuid` assigned, then on a 2nd pass populating the `includedServices`.

chipweinberger:
@nebkat , I added it back: https://github.com/chipweinberger/flutter_blue_plus/commit/a42ca5dd5f1a933b4ce124f0150ecd1a0576aad8

I just temporarily removed it.

 The underlying representation changed.

Now instead of returning a `List<List<BmBluetoothService>>`, we now return `List<BmBluetoothService>` + added a `primaryServiceUuid` field

chipweinberger:
I'm thinking about your change more.

I think you're right. Instead of an "index" we should probably call it a "handle"

on android this would be the `instanceId`

on ios youre probably right it's better to use the hash. but we could also make our own simple hash too. that's probably better.

chipweinberger:
I'm also wondering if we should remove `primaryServiceUuid` from the public API.

tbh I'm not sure the best way to represent all of this stuff. But I know `secondaryServiceUuid` was not the right solution.

chipweinberger:
looking at this, the call it handle as well

```
@interface CBCharacteristic : NSObject {
    CBUUID * _UUID;
    NSArray * _descriptors;
    NSNumber * _handle;
    BOOL  _isBroadcasted;
    BOOL  _isNotifying;
    CBPeripheral * _peripheral;
    long long  _properties;
    CBService * _service;
    NSData * _value;
    NSNumber * _valueHandle;
}
```

and in the BLE spec as well, they use handles

```
Service UUID: 0x180F (Battery Service)
  - Handle: 0x0001
  - Attribute: Included Service
    - Handle: 0x0002 (reference to Environmental Sensing Service)
    - UUID: 0x181A (Environmental Sensing Service)

Service UUID: 0x181A (Environmental Sensing Service)
  - Handle: 0x0002
  - Characteristics: ...
  ```

**Comment by chipweinberger** [comment]
another idea, what if we instead added a handle to Guid?

```
class Guid {
  final List<int> bytes;
  final String? uniqueHandle;
```

string representation:

```
if (uniqueHandle) {
   // 0000-0000-1000-8000-00805f9b34fb:$uniqueHandle
}
```

**Comment by chipweinberger** [comment]
I like it. However, I would use the same existing names where possible. i.e. characteristicUuid instead of just uuid.

**Comment by chipweinberger** [comment]
but the more a think about it. I think a clean 1.0 version is to not have a `BluetoothAttribute` class.

just add indexes to the existing classes. put them all together not interleaved, looks nicer:

```
  final DeviceIdentifier remoteId;
  final Guid serviceUuid;
  final Guid characteristicUuid;
  final Guid? primaryServiceUuid;

  final int serviceIndex;
  final int characteristicIndex;
  final int? primaryServiceIndex;
```

However, Internally, I still like the idea of using the `uuid:%index` strings. For example, in the Bm* classes, and in the `locateCharacteristic` functions. There's no need to add a bunch more arguments everywhere. It just gets too ugly / complicated.

**Comment by nebkat** [comment]
> I like it. However, I would use the same existing names where possible. i.e. characteristicUuid instead of just uuid.

`BluetoothCharacteristic.characteristicUuid` seems a bit redundant which is why I was thinking a deprecated getter would make sense. Similarly `BluetoothCharacteristic.service.uuid` is easier to read than `BluetoothCharacteristic.serviceUuid` and only costs 1 extra character.

> Or maybe we can go 2.0 if we provide 1.0 compatible constructors as well.

Which of these classes should actually be constructible by the user? If I'm not mistaken none of these characteristics/services will work unless they are discovered in the low level API, so a user should really get a reference to our instance of the `BluetoothCharacteristic` rather than creating their own.

It would be great if we could save a round of development and go straight to 2.0, I don't think any of these changes would involve much migration - if any.

> However, Internally, I still like the idea of using the uuid:$index strings. For example, in the Bm* classes, and in the locateCharacteristic functions. There's no need to add a bunch more arguments everywhere. It just gets too ugly / complicated.

Internally I don't mind, we could even reference everything in the format: `$PRIMARY_SERVICE_UUID:$PRIMARY_SERVICE_INDEX/$SERVICE_UUID:$SERVICE_INDEX/$CHARACTERISTIC_UUID:$CHARACTERISTIC_INDEX/$DESCRIPTOR_UUID`. Would need a little bit more thought but you get the idea. We do exchange a lot of mapped data that is probably unnecessary if we could directly identify the service/char/descriptor. It is also duplicated all over the Java/Objective-C classes.

If you are open to it there are many ways we could significantly simplify the code that communicates between Dart/Java/Objective-C.

> a guid is supposed to be globally unique. So I think it still makes more sense than what we have today. Obviously it was poorly named in the original flutter_blue.

This is a confusing concept within GATT but a UUID is meant to uniquely identify the _type_ of the data, _not_ the characteristic itself. If we attempt to correct this by changing what a Guid is (which is just another name for UUID) we will add to the further confusion - instead I think we should educate users on the distinction.

**Comment by chipweinberger** [comment]
All fair. Just want it to be 1.0 api compatible, with no more than ~1 small breaking change.

Any suggestions?

**Comment by nebkat** [comment]
> All fair. Just want it to be 1.0 api compatible, with no more than 1 or 2 small breaking changes.
>
> Any suggestions? I think modifying Guid is the easiest way to support this feature in 1.0 friendly way.

Let me make a PR this weekend, I think it will be possible with 0 or 1 breaking changes.

> not necessarily. If it was just elm.uuid, it's no longer clear. its a minor detail. But I have no plans to deprecate characteristicUuid. Having both is better. In fact, the [original code](https://github.com/pauldemarco/flutter_blue/blob/master/lib/src/bluetooth_characteristic.dart) just used uuid, but I know from experience - it got annoying fast.

If `elm.uuid` is unclear it is `elm` which is the issue - I tend to name my characteristics `heartbeatChar`. But I take your point, can leave `characteristicUuid` as a getter.

> It probably took you about 0.2 seconds to find out. But when it's named characteristicUuid, its 0.0 seconds. Navigating a code base with dozens of instances just like this, becomes annoying fast.

These instances will practically all be eliminated - there will be a single place in the code where we map from the `$SERVICE_UUID:$SERVICE_INDEX/$CHARACTERISTIC_UUID:$CHARACTERISTIC_INDEX` style string to an instance of `BluetoothCharacteristic` and nowhere else internally will care about UUIDs at all.

```

nebkat:
On further investigation a lot of things seems strangely overcomplicated with all of the `BluetoothXXXX` classes acting more like tokens providing views into maps within the main class. Is there a reason why for example a `BluetoothCharacteristic` doesn't maintain it's own value, list of descriptors, etc rather than pulling from `_knownServices`?

chipweinberger:
yes. because users can instantiate these objects themselves

so they need to be able to access the last value from a global store

chipweinberger:
Another thing to keep in mind, verbose logs.

this `$SERVICE_UUID:$SERVICE_INDEX/$CHARACTERISTIC_UUID:$CHARACTERISTIC_INDEX` style string sounds like it would be more annoying for viewing the verbose logs, which a lot of users find important.

nebkat:
This is ending up a slightly bigger change than anticipated but I'll get it finished and you can consider the option. So far looking like a pretty huge reduction in code bloat.

There is possibly a pretty neat solution to not breaking the token style construction with something like:

```dart
  factory BluetoothCharacteristic({
    DeviceIdentifier remoteId,
    Guid this.serviceUuid,
    Guid this.characteristicUuid,
    Guid? primaryServiceUuid,
  }) {
    return FlutterBluePlus.findCharacteristic(remoteId, serviceUuid, primaryServiceUuid, characteristicUuid);
  }
```

> I think adding index to Guid is still the most obvious solution that meets all criteria, without blowing up the code base.

I don't doubt that this is the simplest solution but I think it's important to ensure that implementation details are not leaking into the public API and complicating things in the future. If we were to do this would a `Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", null)` be equivalent to `Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", 1)`?

chipweinberger:
> If we were to do this would a Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", null) be equivalent to Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", 1)?

No. They are not equivalent. 

> serviceUuid = Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", null) 

this refers to the first service with uuid =  `a87ad7d7-b3a9-4a80-a34e-300e7398c92c`

> serviceUuid = Guid("a87ad7d7-b3a9-4a80-a34e-300e7398c92c", 1) 

this refers to the second service with uuid =  `a87ad7d7-b3a9-4a80-a34e-300e7398c92c`

chipweinberger:
> This is ending up a slightly bigger change than anticipated

it's unlikely to be merged. but ill review it.

chipweinberger:
>     return FlutterBluePlus.findCharacteristic(remoteId, serviceUuid, primaryServiceUuid, characteristicUuid);

this is not entirely the same meaning. You can instantiate characteristic that don't exist yet / before discoverServices.

so it would better be called

> FlutterBluePlus.createCharacteristic(remoteId, serviceUuid, primaryServiceUuid, characteristicUuid); 

which in that case you might as well just make it part of the constructor.

---

I'd only want to get rid of the global map if we remove the public constructor.

but again, this feature does not warrant such large changes IMO.

nebkat:
> If users want to compare uuid, they will need to compare guid.uuid instead.
> instead of representing a UUID, GUID now represents a specific svc/chr/desc.

Note that there is no difference in meaning of the term GUID and UUID (globally vs universally) so I am failing to see how this would not confuse users... But to save us going in circles I suggest we get some outside opinions on it if possible.

> this is not entirely the same meaning. You can instantiate characteristic that don't exist yet / before discoverServices.

I may try to represent these as `AnonymousBluetoothCharacteristic` or `BluetoothCharacteristicToken`. IMO this usage should be discouraged over performing service/characteristic discovery as is done on Android/iOS natively.

> it's unlikely to be merged if its a drastic change. but ill review it.

Of course. I understand this will be very difficult to merge, at this stage it is more of an exploration of what could be possible in future.

Currently I have over 600 lines _removed_ on the Dart side alone while maintaining almost identical external API and functionality, plus fully supporting instances of `BluetoothService`/`BluetoothCharacteristic`/`BluetoothDescriptor` that uniquely identify their respective back-end classes.

If you are open to it I think it could be a good development for the library, as I see many parts have not really changed since the original library but code size has ballooned quite a bit.


> this feature is not important enough to make big changes IMO.

Unfortunately our application requires complex usage of BLE so for me it is essential to have these features. It consists of a hub device that is connected by wire to a number of other devices, each of which is represented by a primary service + their respective secondary services i.e. duplicate UUIDs everywhere.

This is for sure an exotic setup but I have personally found that poor library support contributes a lot to such use cases being rare, rather than there not being demand. 

---

In any case thank you for all your responses so far!

chipweinberger:
> If you are open to it I think it could be a good development for the library, as I see many parts have not really changed since the original library but code size has ballooned quite a bit.

I know you don't intend it as an insult, but what you don't see are over ~60 bugs in the original library fixed, some that were design problems. Among a handful of features added, comprehensive error handling, etc.

nebkat:
> I know you don't intend it as an insult, but what you don't see are over ~50 bugs in the original library fixed, some that were design problems. Among a handful of features added, comprehensive error handling, etc.

No not at all, we are very grateful for the improvements from the original!

I am mainly just referring to areas of duplicated code (in particular the invoke method blocks and the `_methodStream` mapping). You can have a look at the WIP so far, I have extracted a lot of this code into functions and mixins:

https://github.com/nebkat/flutter_blue_plus/commit/1ecf36dbdecc4fa00ef05a1ec8173177332a874b

I discovered yesterday that once the index parameter is introduced it is no longer necessary to pass around the parent service UUID, since each service can be uniquely identified by its own uuid/index alone.

Also learned that a primary service can include other primary services, so the presence of a parent primary service does not necessarily make it a secondary service.

Lastly, I have yet to test but I think one possible advantage of using `hash` on iOS may be that it preserves the same value as long as the handle remains the same, whereas index may change if a service/characteristic is injected/removed at an earlier index. This may be important if a services reset occurs to avoid the need to invalidate all existing instances of `BluetoothCharacteristic`.

chipweinberger:
> Also learned that a primary service can include other primary services, so the presence of a parent primary service does not necessarily make it a secondary service.

I dont think this is correct.

nebkat:
> I dont think this is correct. If you have a parent, you're not primary.

From Bluetooth Core Spec v5.4 | Vol 3, Part G | Page 1472:

image

image

So primary → primary is possible, as well as primary → secondary → secondary.

The good news is on Android it appears that all the secondary services are included in `BluetoothGatt.getServices()` - so they can always be uniquely identified by their UUID and their handle without the need for hierarchy information. I presume iOS is the same, will find out hopefully later today.

chipweinberger:
ah okay,

the first thing we should do is to fix the existing code then

and add docs to the codebase with explanations

nebkat:
How about we use the pointer address of the `CBCharacteristic`/`CBService` instances, (optionally together with the hash), to uniquely identify each one? This will be easier than finding the index. Then we flatten the services list with an `NSSet` - so we simply find the service with the matching pointer. When [`didModifyServices`](https://developer.apple.com/documentation/corebluetooth/cbperipheraldelegate/peripheral(_:didmodifyservices:)) gets called we remove from the set.

chipweinberger:
pointer seems like a good idea

except i'm not sure how a user would instantiate it themself then

nebkat:
I think if we allow the `index` parameter to be `null`, and only filter by it if it is not `null`, we will solve 99% of that use case.

So if you do `devices.servicesList.firstWhere((s) => s.uuid == ...).characteristics.firstWhere((c) => c.uuid == ...)` you are given a concrete characteristic that is fully indexed, whereas if you initialize it with `BluetoothCharacteristic(characteristicUuid: ..., serviceUuid: ...)` it's anonymous and will match the first service and characteristic.

If someone needs to do multi-UUID (service or characteristic) setups they should use discovery mode. This is the same logic as [`BluetoothGatt.getService(UUID)`](https://developer.android.com/reference/android/bluetooth/BluetoothGatt#getService(java.util.UUID)).

Indices will break if services are inserted/removed at an earlier position in the table, whereas pointers should remain valid.

chipweinberger:
Okay, sure, lets do ptr.

but keep it null unless there are multiple of the same service, chr, etc.

chipweinberger:
looking reasonable. some early feedback

1. no need for `BluetoothValueAttribute`. It' makes the code harder to understand and modify/diverge in the future. If we needed a single source of truth sure, but in this case, I *want* the code to be able to diverge easily. Since they are really not the same thing.

chipweinberger:
Okay, I kept trying to give more feedback, but the PR is too big for me to wrap my head around right now.

There's a a lot of changes here.

_**A lot of it I do like, and would merge.**_

Now that you have a very good understanding of the FBP code, can you submit PR requests incrementally? A lot of these changes can be their own PR.

chipweinberger:
`AdvertisementData` should not have a platform name. The platform name comes from the platform, not the advertisement. It's not part of the BLE spec.

```
  AdvertisementData({
    required this.advName,
    required this.platformName,
```

chipweinberger:
I find this really hard to read.

`// 00000000-0000-0000-0000-000000000000:12345678/00000000-0000-0000-0000-000000000000:12345678/00000000-0000-0000-0000-000000000000:12345678`

It's easier to read like this:

```
{
 service_uuid: xxxx,
 characteristic_uuid: xxxx,
 descriptor_uuid: xxxx,
}
```

 You might think "internal" representations don't matter. But it's not internal. We log these everywhere. Legibility is super important.

nebkat:
This was an accidental push to the PR as I had a reference to the fork@master, but yeah it's a current snapshot of WIP.

> Okay, I kept trying to give more feedback, but the PR is too big for me to wrap my head around right now.

Yeah sorry it's not very pretty right now but I've had to get it to a working state as I have deadlines with our project.

> Now that you have a very good understanding of the FBP code, can you submit PR requests incrementally? A lot of these changes can be their own PR.

Can do my best to break up changes into their own commits. Maybe if you could comment in the code areas you are 100% happy with can start extracting and rebasing piece by piece until its a smaller diff?

-----

> no need for BluetoothValueAttribute. Too much DRY. I want the code to be able to diverge easily.

Personally suggest maximum DRY until the divergence actually occurs. It's pretty easy to un-abstract if the need arises, but without great care the duplicated code can quickly balloon to a difficult to manage state.

> Since chr & desc are really not the same thing. In fact, I've considered changing desc.value to some other name many times. 

As far as the BLE spec is concerned they are almost identical (see `4.8 CHARACTERISTIC VALUE READ` and `4.12 CHARACTERISTIC DESCRIPTORS`), and the descriptor value is referred to as such. What were your thoughts behind changing it?

> Plus having code that looks for either of the underlying OnXXXExent types is not so clean.

Note that prior to these changes the same functions were re-parsing the events everywhere:

```dart
Stream<List<int>> get lastValueStream => FlutterBluePlus._methodStream.stream
      .where((m) => m.method == "OnCharacteristicReceived" || m.method == "OnCharacteristicWritten")
      .map((m) => m.arguments)
      .map((args) => BmCharacteristicData.fromMap(args))
```

So I figured delegating discrimination to the `.where((e) => e.attribute == this)` was a pretty elegant solution as it's 0 cost.

----

> your changes for this code may be fewer lines, but they are harder to read & understand then before. Keeping the separate hasNotify, hasIndicate lines is easier to read & understand for the next person.
> ```
> bool get isNotifying {
>    List<int> lastValue = cccd?.lastValue ?? [];
>    return lastValue.isNotEmpty && (lastValue[0] & 0x03) > 0;
>  }```

Agree it's rather vague as is, how about `lastValue[0] & (CCCD_BIT_NOTIFY | CCCD_BIT_INDICATE)`?

I do believe as a general principle that brevity like this is beneficial and makes classes more readable:
```dart
bool get isNotifying => (cccd?.lastValue?._firstOrNull ?? 0) & (CCCD_BIT_NOTIFY | CCCD_BIT_INDICATE) > 0
```

----

> AdvertisementData should not have a platform name. BLE advertisments do not have a platform name.

Ok so this should live in `ScanResult` rather than `AdvertisementData` then, must have accidentally put it there as I was reducing reliance on the `Bm***` classes.

-----

> I find this really hard to read.
> `// 00000000-0000-0000-0000-000000000000:12345678/00000000-0000-0000-0000-000000000000:12345678/00000000-0000-0000-0000-000000000000:12345678`

Yeah this is not strictly the final form, was just convenient for now, _however_...

> Mind you, it needs to be performant too. I once tried to pretty format everything, and performance tanked.

The fact that we can't pretty print everything means that you will get `{ service_id: 00000000-0000-0000-0000-000000000000:12345678, characteristic_id: 00000000-0000-0000-0000-000000000000:12345678, descriptor_id: 00000000-0000-0000-0000-000000000000:12345678 }` which isn't exactly great when your screen gets spammed with UUID characters. 

Would you be open to this? 20% less characters but still comparatively quite readable:

`[ S@00000000-0000-0000-0000-000000000000:12345678, C@00000000-0000-0000-0000-000000000000:12345678, D@00000000-0000-0000-0000-000000000000:12345678 ]`

Either way I'm happy with some separation of them to cut out the string parsing on native side.

lib/src/flutter_blue_plus.dart
```diff
@@ -436,6 +410,50 @@ class FlutterBluePlus {
...
+    } else if (call.method == OnDescriptorWrittenEvent.method) {
+      BmDescriptorData r = BmDescriptorData.fromMap(call.arguments);
+      return OnDescriptorWrittenEvent(r);
+    } else if (call.method == OnScanResponseEvent.method) {
+      BmScanResponse r = BmScanResponse.fromMap(call.arguments);
+      return OnScanResponseEvent(r);
+    } else {
+      throw UnimplementedError("methodCallMap: ${call.method}");
+    }
+  }
```
nebkat, line 455: This could be a good place to start, together with `_extractEventStream<T>`.

chipweinberger:
too much to discuss at one time.

let's just do it one PR at a time.

lib/src/flutter_blue_plus.dart
chipweinberger: seems reasonable. Instead of `dynamic` lets use a dart enum.

also, no need for 'if else', just use 'if' here. It's easier on the eyes.

chipweinberger: i dont really like the name.

maybe `getEvents`

nebkat: > seems reasonable. Instead of dynamic lets use a dart enum.

Not sure what you mean by this? At best we could give the event classes a common superclass so it's not dynamic, but we'll still always be performing runtime type testing.

> also, no need for 'if else', just use 'if' here. It's easier on the eyes.

I'll replace with a switch statement.

nebkat:
I think it would be splitting hairs here to suggest that a mixin is somehow significantly less readable than the line it replaced - when that line is copied 2-10x in the file. Could equally argue the mixins allow one to focus on what is unique to each event and really this is the perfect use-case for them.

It is however an issue of maintainability when a simple change in the parameters of a class requires 10 identical changes elsewhere. This introduces a very real risk of missing one of the required changes and makes it more difficult to understand and review changes.

To take a concrete example: https://github.com/chipweinberger/flutter_blue_plus/commit/2064c18e40d1a01e2bf1e382c746acc588cc029f, the line `.where((p) => p.primaryServiceUuid` is added 7 times in this relatively simple change. Would an external contributor remember to put in all 7? From viewing the diff alone it's impossible to assess that.

----

There is of course no single right answer on the degree to which the code should be abstracted, but I can tell you from my perspective as an external contributor that in its current form the library is quite difficult to work with primarily due to the amount of duplicated code. Every time I see a duplicated block I must ask myself why _isn't_ this abstracted if it's doing the exact same thing, is there something unique about it?

In addition to that, while it is merely a metric - when I see files with hundreds or even thousands of lines of code it is very daunting. The 700 lines this branch removed while maintaining the same functionality is 700 less lines I have to understand as a contributor.

chipweinberger:
RE: contributors

I gave up on [flutter_reactive_ble](https://github.com/PhilipsHue/flutter_reactive_ble/tree/master/packages/reactive_ble_mobile/ios/Classes) because it was too DRY, and used too many swift language features that I didn't want to have to learn.

And look where they ended up. They went from the most popular BLE library, to the 2nd most popular. And soon to be the 3rd.

chipweinberger:
anyway, still, lets just go PR by PR, if you want to merge.

nebkat:
> I had to google to remember the rules of mixins.
> mixins: "At first, I found this concept somewhat difficult to understand"

..."until I realized how powerful it was."?

Instead of arguing the merits of not introducing mixins to the codebase because you are not familiar with them, why not learn how they work and benefit from their use? Every language feature is difficult until you learn how to use it.

> readability > "maintainability"

[Maintainability](https://en.wikipedia.org/wiki/Software_maintenance) is no less important than readability. If I am writing professional software I need to have confidence that I can easily develop new features and fix bugs in the library.

> Keeping code "flat" is important.

"Flat" code is primarily about avoiding deeply nested statements. Repeating identical code because it's technically quicker to read in one place vs over multiple classes does not make it flat or readable. Abstractions are useful and necessary - and 

----

> And look where they ended up. They went from the most popular BLE library, to the 2nd most popular. And soon to be the 3rd.

The external API of this library is better and it is more actively maintained, that is why it is popular, not because it lacks abstractions.

----

> anyway, still, lets just go PR by PR, if you want to merge.
> We dont need to agree on everything.

No, we certainly don't. I have obviously been willing to devote a lot of time to these changes and this discussion, and I appreciate the same from you. I'm not adamanant on mixins or any part of this in particular, and I hope we are just getting caught up on minor details.

That said we will need to reach a reasonable consensus if I am going to dedicate further time breaking this up.

chipweinberger:
> Instead of arguing the merits of not introducing mixins to the codebase because you are not familiar with them, why not learn how they work and benefit from their use? Every language feature is difficult until you learn how to use it.

The benefit here is too small.

chipweinberger:
On a higher level, when you are a contributor to a code base, you need to be more receptive to feedback. 

I think we've spent enough time discussing small things.

Let's see how the first PR goes, if you want to try. +1,199 −1,884 is of course too big.

chipweinberger:
"- anytime a notification arrives (characteristics, if subscribed)"

^ This is now more complicated to understand than before. We should just remove the line for descriptors, but keep it for characteristics. Is there some way to do that?

```
abstract class BluetoothValueAttribute extends BluetoothAttribute {
  List<int> _lastValue = [];

  BluetoothValueAttribute({
    required BluetoothDevice device,
    required Guid uuid,
    int? index,
  }) : super(device: device, uuid: uuid, index: index);

  /// this variable is updated:
  ///   - anytime `read()` is called
  ///   - anytime `write()` is called
  ///   - anytime a notification arrives (characteristics, if subscribed)
  ///   - when the device is disconnected it is cleared
  List<int> get lastValue => _lastValue;

  /// this stream emits values:
  ///   - anytime `read()` is called
  ///   - anytime `write()` is called
  ///   - anytime a notification arrives (characteristics, if subscribed)
  ///   - and when first listened to, it re-emits the last value for convenience
  Stream<List<int>> get lastValueStream => 
```

chipweinberger:
I've only given push back for `BluetoothValueAttribute` and the `Mixin`s.

It's really a small amount of pushback, given the size of this PR.

lib/src/flutter_blue_plus.dart
chipweinberger: > not sure what you mean.

Maybe let's use this pattern

```
enum EventType {
	login,
	signup,
	cancel
}

class Event {
	EventType type;
	dynamic underlying;
	LoginEvent get login {
		assert(type == EventType.login);
		assert(underlying is LoginEvent);
		return underlying as! LoginEvent;
	}
}
```

lib/src/bluetooth_attribute.dart
```diff
@@ -0,0 +1,77 @@
...
+    required this.uuid,
+    this.index,
+  });
+
+  DeviceIdentifier get remoteId => device.remoteId;
+
+  @Deprecated('Use remoteId instead')
+  DeviceIdentifier get deviceId => remoteId;
+
+  BluetoothAttribute? get _parentAttribute => null;
```
chipweinberger, line 23: this variable should only exist for services. Not chrs & descs.

nebkat: This was mainly for identifier path generation, so will be removed if we are using a map instead.

lib/src/flutter_blue_plus.dart
nebkat: This introduces the possibility that the `EventType` might not match the underlying class (thus the assertions). Why not just let the consumer directly observe whether `event is LoginEvent`?

The `BluetoothEvents` getters already provide nice typed access to the events if this is for user benefit.

nebkat:
> And it's not really about me. It's about the new Flutter developer learning Dart for the first time.

For your consideration: a developer that is just learning the language is highly unlikely to be poking around the internals of a BLE library, let alone contributing changes. 

> But still, needing to keep these docs generic across descriptors & characteristics is not something I want to care about. There's really no reason for DRY here, since the code/concepts do not need to stay in sync. It just happens to be in sync right now.

It would perhaps help to understand what changes you are anticipating to them - as according to the GATT spec these concepts are very much in sync.

Surely keeping one or two extra lines of documentation in sync is easier than keeping double the code _and_ double the documentation in sync?

----

Anyway, good that we are mostly on the same page then, I'll get started on the events refactor.

chipweinberger:
> It would perhaps help to understand what changes you are anticipating to them - as according to the GATT spec these concepts are very much in sync.

For example, on iOS there are restrictions which descriptors can be written to. These belong in the DESC docs, not CHR docs. And these restrictions can change any time.

chipweinberger:
@nebkat , have you ever read this article? https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

The docs having new "only applies to" conditionals, adding `parentAttribute` to descriptors by accident. It' is exactly the problem this article is describing.

Here its pretty benign. But it quickly can get out of hand, becoming unmaintainable, a mess of complication.

nebkat:
@chipweinberger Yes, I have. It's a good article and it makes a very valid point - however it's important not to take such opinion pieces at face value without considering the specific context.

You seem to be under the impression that avoiding abstraction is implicitly better because there is a _risk_ of it being wrong, which is IMO looking at the problem backwards. _A flat hierarchy should rarely be the goal_. It is one of the possible solutions when direct abstracting doesn't work.

Thus the important question is whether or not a given abstraction is actually _wrong_. If you look at the example given in the article:

> Programmer B feels honor-bound to retain the existing abstraction, but since isn't exactly the same for every case, they alter the code to take a parameter, and then add logic to conditionally do the right thing based on the value of that parameter.
>
> What was once a universal abstraction now behaves differently for different cases.

I absolutely agree that if `BluetoothValueAttribute` itself started exhibiting different **behaviors** depending on its derivation it would cease to be a good abstraction - at which point duplication of the relevant feature is preferrable. But as it stands:
- Functionality is identical between both
- It closely models the a concept that exists in GATT
- There are more lines in common between them than there are in the entire `BluetoothService` class

The single extra line of documentation needed to clarify that notifications are only relevant to characteristics is not a "parameter" as described above. It's totally normal to document different behaviors of derived classes (that's kinda the point of deriving in the first place!). Should [Flex](https://api.flutter.dev/flutter/widgets/Flex/mainAxisAlignment.html) not exist because some of its properties provide extra context for the derived classes?

You also mention the iOS restrictions on writing certain descriptors - but the `BluetoothDescriptor` class still exists and has its own `write` method so I'm not sure how that is relevant?

The point the article is making is that you shouldn't be afraid to remove abstractions when they are no longer serving their purpose. I suggest you read some of the comments on this article, as well as the discussions [here](https://news.ycombinator.com/item?id=37306769).

----

> parentAttribute is now added to descriptors, but not actually used (and should be called parentService)

Figure 6.3: GATT-Based Profile hierarchy from the spec:
image

As you can see service is the parent of characteristic is the parent of descriptor in the GATT hierarchy.

chipweinberger:
That's all fair.

But at the end of the day, it's my project, and I find duplication easier to reason about. 

I don't like abstracted code. It's just my preference.

chipweinberger:
And oh, okay. I see how you intend to use parent now.

I'm not sure if we really need that.

I've thought about adding it before tho. But seems unecessary.

chipweinberger:
> > I dont think this is correct. If you have a parent, you're not primary.
> 
> From Bluetooth Core Spec v5.4 | Vol 3, Part G | Page 1472:
> 
> image
> 
> image
> 
> So primary → primary is possible, as well as primary → secondary → secondary.
> 
> The include is simply a link to another service definition in the table which exists independently. The only difference between primary and secondary, aside from semantics, is that only primary services show up in the `Discover All Primary Services` sub-procedure, whereas secondary have to be manually requested, however the Bluetooth stacks on Android/iOS already handle this for us.
> 
> The good news is on Android it appears that all the secondary services are included in `BluetoothGatt.getServices()` anyway - so they can always be uniquely identified by their UUID and their handle without the need for hierarchy information. I presume iOS is the same, will find out hopefully later today.

it would be great to fix this in the current FBP.

mgeilich:
I really need a solution to supporting multiple characteristics with the same UUID.  This works fine with native Bluetooth on Android and iOS, but not with this plugin.

chipweinberger:
closing. master branch has changed significantly due to the platform interface. this work will need to be redone.

if you attempt it again, please keep the PR small & minimal so that it is easier to review and merge.

PR #1055, opened 2024-11-12
[Feature]: Handle time out in startScan Method

OmarAly92:
This pull request introduces an optional onTimeout parameter to the startScan method. The onTimeout callback function is triggered in the stopScan method when the scan times out, providing user a way to handle timeout events directly.

**Changes**

- Added void Function()? onTimeout as an optional parameter to startScan.
- Updated _scanTimeout in stopScan to invoke onTimeout?.call() after the scan stops due to a timeout.

**Benefit**

This improvement allows developers to handle timeout events with a custom callback, increasing flexibility and control over scan behavior.

**Testing**

I have tested the changes, and the functionality is working as expected. The onTimeout callback is triggered correctly upon timeout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,12 +186,16 @@ FlutterBluePlus.cancelWhenScanComplete(subscription);
 // In your real app you should use `FlutterBluePlus.adapterState.listen` to handle all states
 await FlutterBluePlus.adapterState.where((val) => val == BluetoothAdapterState.on).first;
 
-// Start scanning w/ timeout
+// Start scanning w/ timeout and onTimeout callback if you want to do action after timeout
 // Optional: use `stopScan()` as an alternative to timeout
 await FlutterBluePlus.startScan(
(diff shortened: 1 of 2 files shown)


chipweinberger:
pretty good idea. 

currently you would use `isScanning` to do this

OmarAly92:
Making timeout manual is a good approach and working well but adding it inside package makes it easier for users to handle timeouts without extra setup

chipweinberger:
there is a problem

```
            await stopScan();
            onTimeout?.call();
```

if `stopScan` throws exception, then `onTimeout` will not be called.

chipweinberger:
also another problem: if the user calls `stopScan` manually, the `onTimeout` will not be called.

chipweinberger:
i will not merge this. I suggest you use `FlutterBluePlus.isScanning` to detect when scanning stops.

PR #1067, opened 2024-12-01
Using uint8 internally to get better performance

evan361425:
Detailed discussion in #1010

Tested on Android system, but not iOS.


chipweinberger:
"buildPerformanceButton"

what does this screen look like?

lib/src/bluetooth_msgs.dart
```diff
@@ -127,27 +127,19 @@ class BmScanAdvertisement {
 
   factory BmScanAdvertisement.fromMap(Map<dynamic, dynamic> json) {
     // Get raw data
-    var rawManufacturerData = json['manufacturer_data'] ?? {};
-    var rawServiceData = json['service_data'] ?? {};
-    var rawServiceUuids = json['service_uuids'] ?? [];
+    Map<Object?, Object?> rawManufacturerData = json['manufacturer_data'] ?? <int, List<int>>{};
+    Map<Object?, Object?> rawServiceData = json['service_data'] ?? <int, List<int>>{};
+    List<Object?> rawServiceUuids = json['service_uuids'] ?? <String>[];
```
chipweinberger, line 132: please use `var`. Or give it a proper type. not Object.

lib/src/flutter_blue_plus.dart
```diff
@@ -746,7 +746,7 @@ class ServiceDataFilter {
   // The 'mask' must have the same length as 'data'.
   List<int> mask;
 
-  ServiceDataFilter(this.service, {this.data = const [], this.mask = const []});
+  ServiceDataFilter(this.service, {required this.data, required this.mask});
```
chipweinberger, line 749: this is a breaking change. please undo it.

lib/src/flutter_blue_plus.dart
```diff
@@ -726,7 +726,7 @@ class MsdFilter {
   /// The 'mask' must have the same length as 'data'.
   List<int> mask;
 
-  MsdFilter(this.manufacturerId, {this.data = const [], this.mask = const []});
+  MsdFilter(this.manufacturerId, {required this.data, required this.mask});
```
chipweinberger, line 729: this is a breaking change. please undo it.

chipweinberger:
1. have you tested it an iOS
2. please improve the "log" function so that we still log in the same format as before (unformatted hex string)

lib/src/bluetooth_msgs.dart
evan361425: rollback as before

lib/src/flutter_blue_plus.dart
evan361425: 👌

lib/src/flutter_blue_plus.dart
evan361425: 👌

evan361425:
- d1df6df rollback breaking changes and perf screen
- 49a808b log hex when getting uint8list

evan361425:
help needed for iOS device, can you help me test it?

chipweinberger:
thanks for the update. I can't test it right now, but maybe another day.

chipweinberger:
would need rebase

PR #1091, opened 2025-01-03
[Feature] add platform interface

### LINKED ISSUE #769: [Feature]: Add Support for Web Bluetooth W3C standard ( preliminary )
### FlutterBluePlus Version

all

### Flutter Version

all

### What OS?

All

### OS Version

Supported Chrome

### Bluetooth Module

any

### What is your feature request?

With Chrome ( behind flag ) supporting Web Bluetooth standard, This can be a great addition for Flutter Web Development.

If anyone is interested please let me know

### Logs

```shell
Chrome Web BLE support:
https://developer.chrome.com/docs/capabilities/bluetooth?hl=en

Web Bluetooth W3C Spec:
https://webbluetoothcg.github.io/web-bluetooth/
```

### LINKED ISSUE #983: [Feature]: Linux Support
### FlutterBluePlus Version

all

### Flutter Version

all

### What OS?

All

### OS Version

all

### Bluetooth Module

all

### What is your feature request?

Add Linux support to FBP

### Logs

```shell
h
```

tnc1997:
Closes #769, Closes #983


chipweinberger:
feel free to take the initiative

chipweinberger:
Hey Thomas, 

I've been thinking about a new semi-federated approach, and just wanted to give you a heads up & opportunity to weigh in, if you are still interested.

Basically, federated has made the release process too complicated & unenjoyable. I loath doing them. I think there is a better hybrid approach!

Here's my motivation:
* federated releases take me 5-10 minutes, when they used to take 30 seconds.
* it's really easy to mess up a federated releases, because of the dependency graph. Gotta release in the right order & make sure like 6 different files are all updated correctly. e.g. https://github.com/chipweinberger/flutter_blue_plus/commit/55284978b1cd3ffa4eea9b80db0ea0ebf145de35 & if you mess up you gotta do even more releases!
* maintaining 6 separate changelogs is tiresome. a unified changelog is simpler for everyone.
* the complexity of supporting people mixing & matching various FBP packages is not worth the benefit, especially when it's because a user just forgot to update their platform dependency
* dev work is more annoying since you need to set multiple packages to dev (outside of the Example app)

It's pretty clear there is a decent amount of complexity to federated. There is a real cost to the federated approach.

But what about the benefits?
1. users can implement their own platform implementations

That's a pretty cool benefit!

Well, I think we can have the best of both worlds! Just put everything in the same package, but keep the federated architecture.

1. rename `/packages` to `/components` & put everything into the main FlutterBluePlus package
2. add a new `registerPlatform` function to FBP

By default we would register all 5 of the platforms with our own implementations, but users could also provide their own and override the default one. It also lets people easily swap in a "debug" / "simulator" implementation for testing, dev, etc, too.

Versioning & compatibility could be checked too, on registration.

_Yes, there is a "cost" to doing things differently._ The main "limitation" of this approach is that compatibility checks are now hybrid runtime + compile time, instead of just compile time during build. But that's pretty minor to me, given the advantage of simplifying the releases, dev, support, and changelogs.

Anyway, these are just my thoughts and I'm fairly convinced this is the approach I'll take going forward, just wanted to let you know!

Kind regards,
Chip

tnc1997:
Hi @chipweinberger 

Personally I think that it makes sense to follow the federated plugin architecture that Flutter has standardised as this is what developers are familiar with amongst other reasons but you are of course welcome to continue development of Flutter Blue Plus and the associated platform implementations as you wish.

Cheers, Thomas.

PR #1109, opened 2025-01-19
Feature/duplicateuuids

mgeilich:
Added support for multiple characteristics within a service with the same UUIDs.  This solution adds a property, characteristicId, to characteristic data structures on the Flutter side, to disambiguate the characteristics.  These are passed through the methodChannel to the native side, where the new property is used as an index into the array of characteristics for the service, so that the correct one of the duplicates is used in bluetooth operations.

This solution does not change the FBP API.  The code is written for Android, iOS, Linux, and web but only tested under Android and iOS.


packages/flutter_blue_plus/example/lib/screens/bluetooth_off_screen.dart
```diff
@@ -47,6 +47,13 @@ class BluetoothOffScreen extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    late bool isAndroid = false;
+    try {
+      if (Platform.isAndroid) isAndroid = true;
+    } catch (e) {
+      isAndroid = false;
+    }
```
chipweinberger, line 55: why?

packages/flutter_blue_plus/example/lib/screens/scan_screen.dart
```diff
@@ -27,7 +27,7 @@ class _ScanScreenState extends State<ScanScreen> {
   void initState() {
     super.initState();
 
-    _scanResultsSubscription = FlutterBluePlus.scanResults.listen((results) {
+    _scanResultsSubscription = FlutterBluePlus.onScanResults.listen((results) {
```
chipweinberger, line 30: why?

.vscode/launch.json
```diff
@@ -0,0 +1,26 @@
...
+            "program": "packages/flutter_blue_plus/example/lib/main.dart",
+            "args": [
+            "-d",
+            "chrome",
+            ]
+        },    ],
+    "compounds": [
+
+    ]
+}
```
chipweinberger, line 26: do we need this?

chipweinberger:
there are too many extra changes....

packages/flutter_blue_plus/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -487,7 +487,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = Y2XUKL28VD;
+				DEVELOPMENT_TEAM = WFBC77A558;
```
ekuleshov, line 490: @chipweinberger on a side note, perhaps the `DEVELOPMENT_TEAM` should not be in here.

packages/flutter_blue_plus/example/lib/screens/bluetooth_off_screen.dart
ekuleshov: That is probably an odd way to fix the web issue. Should not be needed anymore with this change:

https://github.com/chipweinberger/flutter_blue_plus/commit/eb15c5b8507a891f4fc645022d1ec8606ae2156e#diff-cd67edaa5b79faf4613c94dff01463569d23d4e58b0a60116b432fb3960788a0R37

ekuleshov:
> there are too many extra changes...

@chipweinberger I agree with this.

Though some of these probably due to auto-format done by an IDE.
Perhaps we should run the dart format on all dart code?

packages/flutter_blue_plus_web/pubspec.yaml
```diff
@@ -10,7 +11,8 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_blue_plus_platform_interface: ^1.0.0
+  flutter_blue_plus_platform_interface:
+    path: ../flutter_blue_plus_platform_interface
```
ekuleshov, line 15: This and the `publish_to: none` above should not be needed anymore if you are using example app for debugging. There are dependency overrides with relative paths in the example app pubspec.

chipweinberger:
btw, there can also be duplicate services.

so that's why imo let's just add an index to Guid

ekuleshov:
> imo let's just add an index to Guid

@chipweinberger wouldn't that break Guid comparison? E.g. when you need to find them all.

Are you trying to avoid adding an `index` field to all request and response classes, including the internal `BmBluetooth*` ones?

chipweinberger:
it would fix comparison.

ekuleshov:
> it would fix comparison.

I didn't realize the Guid comparison is using `str128.hashCode` (could have collisions)...
Perhaps it should have been using `str128` instead...
 
> these characteristics are distinct and should not evaluate as identical

Characteristics - yes. But Guids are not the identity of characteristic, they are more like their types.

Such index inside a Guid feels really odd as it does not correlate with anything in the BLE/GATT specs.
And Guids in advertise data may not match with the service's Guids.

> If you just want to compare the raw 128bits id, you can still do that

I was coming from a common case when you discover services on a device.
You get list of services, each having list of characteristics. Usually expand-filter on service Guid + char Guid using static constants for these. With indexes would have to compare on Guid's `str128`...

chipweinberger:
> I didn't realize the Guid comparison is using str128.hashCode

me neither. we should fix that.

> I was coming from a common case

the common case will still work. If the index is 0, it is ignored / backwards compatible.

> And Guids in advertise data may not match with the service's Guids.

I think thats what you would want. The advertising Service Guid will always have index = null. So if you then go through the discoverServices result, it would match with the first service. Seems very reasonable.

ekuleshov:
> The advertising Service Guid will always have index = null. So if you then go through the discoverServices result, it would match with the first service. Seems very reasonable.

That seem to be a problem. Form a common sense it should match Guids with same str128 and any index.

Generally the Guild with an index is probably not something that should be in the consumer's API. It feels more like something that is internal between FBP Dart and native layers.

E.g. methods on `BluetoothCharacteristic` like `.read()`, `.write()`, `.setNotifyValue()` have access to internal state and perhaps they could handle transforming guid's and/or indexes internally when communicating with native?

chipweinberger:
I think it needs to be in the public facing API, since they need to know what index they are talking to.

ekuleshov:
> I think it needs to be in the public facing API, since they need to know what index they are talking to.
> 
> If a user designs a service with 10 duplicate chrs, they would want to be able to send & receive data to the index of their choosing. I'm not sure how else they would refer to them in a meaningful way, if not by index.

E.g. `BluetoothDevice.servicesList[index]` and `BluetoothService.characteristics[index]` - then work with the instances.

Also, the order may not be defined at all. For example, when ios/macos/android is acting as a peripheral, you'll see the host's OS services + all services custom peripheral adds using host's peripheral API. You can't override the platform ones and can only add to them, so easily can be duplicate services.

chipweinberger:
> That seem to be a problem. Form a common sense it should match Guids with same str128 and any index.

No, because Guid represents both UUID + Index. And it default to index 0. So by common sense it should only match the first one.

ekuleshov:
> No, because Guid represents both UUID + Index.
> And it default to index 0. So by common sense it should only match the first one.

So, far they all been just uuid and those uuids matched what various Bluetooth specs declared. E.g. having a "name" or "battery"  char uuid as in BLE specs becoming uuid+index looks very very odd.

chipweinberger:
Whoever designed BLE this way is so annoying.

> E.g. BluetoothDevice.servicesList[index] and BluetoothService.characteristics[index] - then work with the instances.

Sure, but now they can't pass around just the chrs and know what they refer to. In theory, we could add a "index" convenience function that loops through the array and returns the index. 

> Also, the order may not be defined at all. For example, when ios/macos/android is acting as a peripheral, you'll see the host's OS services + all services custom peripheral adds using host's peripheral API. You can't override the platform ones and can only add to them, so easily can be duplicate services.

Not sure I follow. But lets just focus on Central. Yes Android is more annoying. But again, if you have a device with 1 service and 10 duplicate Chrs, android must have a way to help you talk to the first one, I assume. Otherwise you'll just be guessing and checking.

chipweinberger:
Before we decide, we should probably figure out how each platform handles duplicate CHRs.

1. iOS - respects order
2. Android - exposes handle + *maybe* respects order
3. other - unknown

ekuleshov:
> Sure, but now they can't pass around just the chrs and know what they refer to.

Do they need to? They already got the thing they wanted. And that thing could handle indexes or what's not internally.

> But lets just focus on Central...

I was describing a real life example what central would see from such device. Though mostly the duplicated services. :)

chipweinberger:
> Do they need to? They already got the thing they wanted. And that thing could handle indexes or what's not internally.

Presumably they would like to know which CHR they are referring to, in terms of index.

chipweinberger:
Although I did just think of another complication - service changes.

If your device does a service change and all the services & chrs get reordered, then things will get annoying fast.

Maybe we need both opaque handle + index.

chipweinberger:
I think you're right. Let's just have an opaque handle.

Annoyingly, we will also need to remove the `BluetoothCharacterstic` public constructor, since users will not be able to construct them themselves anymore. But I think that's an acceptable tradeoff.

ekuleshov:
> Presumably they would like to know which CHR they are referring to, in terms of index.
> 
> Put yourself in the shoes of someone who has 1 service and 10 duplicate CHRs. Things are going to get annoying _really fast_ if you can't just always refer to it by index.

Still don't see a problem:

```
var char5 = service[5];
var char7 = service[7];
```

Then you can store these or refer to them any way you like.

> Although I did just think of another complication - service changes.
> If your device does a service change and all the services & chrs get reordered, then things will get annoying fast.
> Maybe we need both opaque handle + index.

Could you please point me to Android API reference for these opaque handles?

chipweinberger:
https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#getInstanceId()

chipweinberger:
I dont want to blow up the code base to support this minor feature, which is why i like the Guid approach.

We could just remove the Guid equality function, and force users to choose how to compare.

mgeilich:
Thanks for the energy on this.  Handling chararacteristics with duplicate UUIDs is part of the Bluetooth spec, an important feature, and essential to the success of my product.  I found my solution not to take a lot of code or time, but I'm not partial to it.  I just need something that works.

My rationale for using an index internally and not changing the external API is that this parallels how our existing code for a native java bluetooth interface works.  Characteristics with duplicate UUIDs are identified by the order in which they are returned, which is consistent.  The first one returned is chr #1, the second is chr #2, etc.

Like I said, not a lot of code or time, can we just get this done one way or the other?

ekuleshov:
@mgeilich could you please merge from main. there are bunch of conflicts (mostly on things that should not have been changed)

chipweinberger:
sure we can merge this. lets fix up the unecessary changes, and rename `characteristicId` -> `characteristicIndex`

mgeilich:
Yes, I can do whatever is necessary to make this acceptable.  Renaming characteristicId is easy.  I am not particularly git-savvy and don't understand "merge from main" yet.  Hopefully you can give me some guidance while I deal with the renaming.  Also, I managed to get the web side of this working and tested.

ekuleshov:
> I am not particularly git-savvy and don't understand "merge from main" yet.

The Github web UI for your PR you see "This branch has conflicts that must be resolved" section and it has instructions for the command line.

Personally I do this in Android Studio/IntelliJ - just pull from the remote origin/master and the IDE will ask you to resolve conflicts with your local branch changes.

After you resolve these conflicts you can just push your branch again and it will update this PR.

mgeilich:
I have merged origin/master into my branch and resolved conflicts.  And done the renaming of characteristicId.  And pushed all of that.  There are build conflicts with the Darwin branch, I'm assuming that's expected?

packages/flutter_blue_plus/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -470,6 +470,10 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
+<<<<<<< HEAD
+				DEVELOPMENT_TEAM = WFBC77A558;
+=======
+>>>>>>> upstream/master
```
ekuleshov, line 476: @mgeilich it doesn't look like you resolved the conflicts.

Though this file should not be changed in your PR. Please revert.

mgeilich: Done, sorry I missed that.  BTW, this merge undoes the solution for web, which required updates to flutter_blue_web.  I guess that will be an independent excercise.

ekuleshov: @mgeilich haven't reverted entire file. see change files in the PR

.vscode/launch.json
ekuleshov: @mgeilich please remove/revert this file as well

packages/flutter_blue_plus/example/lib/screens/scan_screen.dart
ekuleshov: @mgeilich this also need to be reverted or provide some reationale for the change

packages/flutter_blue_plus/lib/src/bluetooth_characteristic.dart
```diff
@@ -10,19 +10,22 @@ class BluetoothCharacteristic {
...
   final Guid serviceUuid;
   final Guid characteristicUuid;
+  final int characteristicIndex;
   final Guid? primaryServiceUuid;
 
   BluetoothCharacteristic({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
ekuleshov, line 20: @chipweinberger should this be an optional parameter with value defaulted to 0? Like `this.characteristicIndex = 0,`

packages/flutter_blue_plus/lib/src/bluetooth_events.dart
```diff
@@ -152,6 +141,7 @@ class OnCharacteristicReceivedEvent {
   BluetoothCharacteristic get characteristic => BluetoothCharacteristic(
       remoteId: _response.remoteId,
       characteristicUuid: _response.characteristicUuid,
+      characteristicIndex: 0,
```
ekuleshov, line 144: @mgeilich should `BmCharacteristicData` also have an index filed and passed it here?

packages/flutter_blue_plus/pubspec.yaml
```diff
@@ -2,6 +2,7 @@ name: flutter_blue_plus
 description: Flutter plugin for connecting and communicating with Bluetooth Low Energy devices.
 version: 1.35.1
 homepage: https://github.com/chipweinberger/flutter_blue_plus
+publish_to: none
```
ekuleshov, line 5: @mgeilich please revert this

packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
```diff
@@ -1310,6 +1327,7 @@ - (void)centralManager:(CBCentralManager *)central
 // ██   ██  ██       ██       ██       ██    ██  ██   ██     ██     ██               
 // ██████   ███████  ███████  ███████   ██████   ██   ██     ██     ███████ 
 
+int characteristicCounter = 0;
```
ekuleshov, line 1330: @mgeilich does it have to be a field? isn't it being reset when characteristics are populated?

packages/flutter_blue_plus_ios/pubspec.yaml
ekuleshov, line 1: @mgeilich please revert this file. the ios project does not exist anymore

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -317,6 +312,7 @@ class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
     return true;
   }
 
+  int characteristicCounter = 0;
```
ekuleshov, line 315: this should be a local variable inside `discoverServices()`

packages/flutter_blue_plus_linux/pubspec.yaml
```diff
@@ -11,7 +12,8 @@ dependencies:
   bluez: '>=0.7.8 <0.9.0'
   flutter:
     sdk: flutter
-  flutter_blue_plus_platform_interface: ^1.0.0
+  flutter_blue_plus_platform_interface:
+    path: ../flutter_blue_plus_platform_interface
```
ekuleshov, line 16: @mgeilich please revert all changes in this file

packages/flutter_blue_plus_platform_interface/lib/flutter_blue_plus_platform_interface.dart
```diff
@@ -6,7 +6,7 @@ export 'src/guid.dart';
 export 'src/log_level.dart';
 
 /// The interface that implementations of flutter_blue_plus must implement.
-abstract base class FlutterBluePlusPlatform {
+abstract class FlutterBluePlusPlatform {
```
ekuleshov, line 9: @mgeilich please revert this change or provide some rationale

pubspec.yaml
ekuleshov, line 1: @mgeilich please revert this

packages/flutter_blue_plus_platform_interface/lib/flutter_blue_plus_platform_interface.dart
mgeilich: My local copy of this line looks like:
abstract class FlutterBluePlusPlatform {

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -318,13 +319,19 @@ class BmBluetoothCharacteristic {
     // convert descriptors
     List<BmBluetoothDescriptor> descs = [];
     for (var v in json['descriptors']) {
-      descs.add(BmBluetoothDescriptor.fromMap(v));
+      final Map<dynamic, dynamic> bv = {};
+      for (var key in v.keys) {
+        bv[key] = v[key];
+        bv['characteristic_index'] = json['characteristic_index'];
```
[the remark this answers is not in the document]
mgeilich: That's true, moved outside the loop...

pubspec.yaml
mgeilich: ?

packages/flutter_blue_plus_platform_interface/lib/flutter_blue_plus_platform_interface.dart
ekuleshov: @mgeilich unfortunately we can't use your local copy and can't see what your change are.

You can see what we see in this PR by looking at this git web UI. Please scroll through your changes and revert/remove any code that is not related to the issue (e.g. platform project files, pubspecs, etc).

https://github.com/chipweinberger/flutter_blue_plus/pull/1109/files

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -132,6 +132,7 @@ class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
     return true;
   }
 
+  int characteristicCounter = 1;
```
ekuleshov, line 135: @mgeilich this should be a local variable inside `discoverServices()` method

mgeilich: Made corresponding scoping changes in other platform modules as well.

packages/flutter_blue_plus/example/lib/screens/scan_screen.dart
mgeilich: reverted...

mgeilich:
Hopefully, all issues are addressed?

packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -2262,7 +2273,9 @@ public void onServicesDiscovered(BluetoothGatt gatt, int status)
             List<Object> services = new ArrayList<Object>();
             for(BluetoothGattService s : gatt.getServices()) {
                 services.add(bmBluetoothService(gatt.getDevice(), s, null, gatt));
+                characteristicCounter = 0;  //reset
                 for(BluetoothGattService s2 : s.getIncludedServices()) {
+                characteristicCounter = 0;  //reset
```
ekuleshov, line 2278: @mgeilich what does this set to 0 in the for loop do?

packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -2639,6 +2653,7 @@ HashMap<String, Object> bmBluetoothService(
         return map;
     }
 
+    int characteristicCounter = 0;
```
ekuleshov, line 2656: @mgeilich shouldn't need to use instance field

ekuleshov:
> Hopefully, all issues are addressed?

Unfortunately not. Please scroll through changed files at https://github.com/chipweinberger/flutter_blue_plus/pull/1109/files

There are still unrelated changes you need to revert.
Also not all comments are addressed or replied to.

packages/flutter_blue_plus/example/ios/Runner.xcodeproj/project.pbxproj
mgeilich: Yes, I have removed this.

packages/flutter_blue_plus/lib/src/bluetooth_events.dart
mgeilich: This solution uses the index to perform operations.  The returned data structures do not contain the index.

packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
mgeilich: I do not see this in the latest merge/commit.

packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
```diff
@@ -126,16 +130,20 @@ class FlutterBluePlus {
...
 
   /// Gets the current state of the Bluetooth module
   static Stream<BluetoothAdapterState> get adapterState async* {
     // get current state if needed
     if (_adapterStateNow == null) {
-      var result = await _invokeMethod(() => FlutterBluePlusPlatform.instance.getAdapterState(BmBluetoothAdapterStateRequest()));
+      if (kIsWeb) {
+        yield BluetoothAdapterState.on;
+        return;
+      }
```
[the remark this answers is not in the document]
tnc1997: @ekuleshov unimplemented methods now return sensible defaults instead of throwing an error.

packages/flutter_blue_plus_platform_interface/lib/flutter_blue_plus_platform_interface.dart
mgeilich: I have reverted this

mgeilich:
Thanks for the link to the remaining issues.  I think I have addressed them all now.

packages/flutter_blue_plus/example/pubspec.lock
ekuleshov, line 1: @mgeilich please revert this change too

packages/flutter_blue_plus/example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -674,6 +674,10 @@
...
```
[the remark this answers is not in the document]
mgeilich: sorry about that.  there were multiple instances, I missed the second one.

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -568,9 +564,8 @@ final class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
           errorString: '',
         ),
       );
-
-      return true;
```
tnc1997, line 572: Could we continue to return `true` if we read the characteristic successfully?

I am unsure as to why this has changed but I may be missing something?

Could the changes to this method be reverted if they are not required?

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -660,6 +655,7 @@ final class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
           remoteId: request.remoteId,
           serviceUuid: request.serviceUuid,
           characteristicUuid: request.characteristicUuid,
+          characteristicIndex: 0,
```
tnc1997, line 658: Should this be the `characteristicIndex` from the `request` instead of `0`?

```suggestion
          characteristicIndex: request.characteristicIndex,
```

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -946,6 +942,7 @@ final class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
           characteristicUuid: Guid.fromBytes(
             characteristic.uuid.value,
           ),
+          characteristicIndex: 0,
```
tnc1997, line 945: Should this be the `characteristicIndex` from the `request` instead of `0`?

```suggestion
          characteristicIndex: request.characteristicIndex,
```

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
tnc1997, line 1: When reading/writing characteristics/descriptors should we take the `characteristicIndex` into consideration?

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -300,15 +300,16 @@ class BmBluetoothCharacteristic {
...
   final Guid? primaryServiceUuid;
   List<BmBluetoothDescriptor> descriptors;
   BmCharacteristicProperties properties;
 
-
   BmBluetoothCharacteristic({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 312: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -336,13 +343,15 @@ class BmBluetoothDescriptor {
...
   final Guid characteristicUuid;
+  final int characteristicIndex;
   final Guid descriptorUuid;
   final Guid? primaryServiceUuid;
 
   BmBluetoothDescriptor({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 354: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -439,12 +449,14 @@ class BmReadCharacteristicRequest {
...
   final Guid serviceUuid;
   final Guid characteristicUuid;
+  final int characteristicIndex;
   final Guid? primaryServiceUuid;
 
   BmReadCharacteristicRequest({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 459: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -642,6 +636,7 @@ final class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
...
```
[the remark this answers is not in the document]
mgeilich: No, the index is only used for initiating operations on the correct characterist.  It is not returned in the returned data structures.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -499,13 +511,15 @@ class BmReadDescriptorRequest {
...
   final Guid characteristicUuid;
+  final int characteristicIndex;
   final Guid descriptorUuid;
   final Guid? primaryServiceUuid;
 
   BmReadDescriptorRequest({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 522: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -531,16 +546,17 @@ class BmWriteCharacteristicRequest {
...
   final BmWriteType writeType;
   final bool allowLongWrite;
   final List<int> value;
 
-
   BmWriteCharacteristicRequest({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 559: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -633,15 +656,16 @@ class BmSetNotifyValueRequest {
...
   final Guid? primaryServiceUuid;
   final bool forceIndications;
   final bool enable;
 
-
   BmSetNotifyValueRequest({
     required this.remoteId,
     required this.serviceUuid,
     required this.characteristicUuid,
+    required this.characteristicIndex,
```
tnc1997, line 668: Should we default this to `0` and make it optional in order to avoid a breaking change?

```suggestion
    this.characteristicIndex = 0,
```

https://github.com/chipweinberger/flutter_blue_plus/issues/795#issuecomment-2461566048

ekuleshov:
@mgeilich I checked out branch you created this PR from at https://github.com/mgeilich/flutter_blue_plus/tree/feature/duplicateuuids

And I'm getting bunch of dart errors and lints.

image

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
mgeilich: Same comment as previous.  The index is only used for initiating operations.  It is not returned.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -318,13 +319,19 @@ class BmBluetoothCharacteristic {
...
+      }
+      bv['characteristic_index'] = json['characteristic_index'];
+      descs.add(BmBluetoothDescriptor.fromMap(bv));
     }
 
     return BmBluetoothCharacteristic(
       remoteId: DeviceIdentifier(json['remote_id']),
       serviceUuid: Guid(json['service_uuid']),
       characteristicUuid: Guid(json['characteristic_uuid']),
+      characteristicIndex: int.parse(json['characteristic_index']),
```
tnc1997, line 334: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
      characteristicIndex: json['characteristic_index'],
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -352,6 +361,7 @@ class BmBluetoothDescriptor {
...
```
tnc1997, line 364: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
      characteristicIndex: json['characteristic_index'],
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
mgeilich: we're not trying to pass it back.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -453,6 +465,7 @@ class BmReadCharacteristicRequest {
     data['remote_id'] = remoteId.str;
     data['service_uuid'] = serviceUuid.str;
     data['characteristic_uuid'] = characteristicUuid.str;
+    data['characteristic_index'] = characteristicIndex.toString();
```
tnc1997, line 468: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
    data['characteristic_index'] = characteristicIndex;
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -515,6 +529,7 @@ class BmReadDescriptorRequest {
...
```
tnc1997, line 532: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
    data['characteristic_index'] = characteristicIndex;
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
mgeilich: could be, but I went for keeping the changes localized instead of widespread.  I don't have a use case for using it with descriptors.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -552,6 +568,7 @@ class BmWriteCharacteristicRequest {
...
```
tnc1997, line 571: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
    data['characteristic_index'] = characteristicIndex;
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -583,6 +602,7 @@ class BmWriteDescriptorRequest {
     data['remote_id'] = remoteId.str;
     data['service_uuid'] = serviceUuid.str;
     data['characteristic_uuid'] = characteristicUuid.str;
+    data['characteristic_index'] = characteristicUuid.toString();
```
tnc1997, line 605: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
    data['characteristic_index'] = characteristicUuid;
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -652,6 +676,7 @@ class BmSetNotifyValueRequest {
     data['remote_id'] = remoteId.str;
     data['service_uuid'] = serviceUuid.str;
     data['characteristic_uuid'] = characteristicUuid.str;
+    data['characteristic_index'] = characteristicIndex.toString();
```
tnc1997, line 679: Could we use an `int` instead of a `String` when serializing and deserializing the index?

```suggestion
    data['characteristic_index'] = characteristicIndex;
```

https://docs.flutter.dev/platform-integration/platform-channels#codec

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -161,23 +161,31 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
+          // Still have to add the characteristic even if this fails
+          int characteristicCounter = 0;
+          try {
+            for (final d in (await c.getDescriptors().toDart).toDart) {
+              descriptors.add(
+                BmBluetoothDescriptor(
+                  remoteId: device.remoteId,
+                  serviceUuid: Guid.fromString(s.uuid),
+                  characteristicUuid: Guid.fromString(c.uuid),
+                  characteristicIndex: characteristicCounter,
```
tnc1997, line 173: Should we be incrementing the `characteristicCounter` on each iteration of the for loop?

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -265,12 +272,10 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
 
-      final service =
-          await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
+      final service = await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
 
-      final characteristic = await service
-          .getCharacteristic(request.characteristicUuid.str128.toJS)
-          .toDart;
+      final characteristics = await service.getCharacteristics().toDart;
+      final characteristic = characteristics[request.characteristicIndex];
```
tnc1997, line 278: Should we check that the `characteristicIndex` is valid given the length of `characteristics`?

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -394,12 +396,10 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
 
-    final service =
-        await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
+    final service = await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
 
-    final characteristic = await service
-        .getCharacteristic(request.characteristicUuid.str128.toJS)
-        .toDart;
+    final characteristics = await service.getCharacteristics().toDart;
+    final characteristic = characteristics[request.characteristicIndex];
```
tnc1997, line 402: Should we check that the `characteristicIndex` is valid given the length of `characteristics`?

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -327,16 +332,11 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
       }
 
-      final service =
-          await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
+      final service = await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
 
-      final characteristic = await service
-          .getCharacteristic(request.characteristicUuid.str128.toJS)
-          .toDart;
+      final characteristic = await service.getCharacteristic(request.characteristicUuid.str128.toJS).toDart;
```
tnc1997, line 337: Should this be updated to use the `characteristicIndex` in the `request` similar to the other methods?

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -545,21 +544,15 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
 
-      final service =
-          await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
+      final service = await gatt.getPrimaryService(request.serviceUuid.str128.toJS).toDart;
 
-      final characteristic = await service
-          .getCharacteristic(request.characteristicUuid.str128.toJS)
-          .toDart;
+      final characteristics = await service.getCharacteristics().toDart;
+      final characteristic = characteristics[request.characteristicIndex];
```
tnc1997, line 550: Should we check that the `characteristicIndex` is valid given the length of `characteristics`?

packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart
```diff
@@ -615,26 +608,21 @@ final class FlutterBluePlusWeb extends FlutterBluePlusPlatform {
...
```
tnc1997, line 614: Should we check that the `characteristicIndex` is valid given the length of `characteristics`?

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
tnc1997: If understood correctly we use `characteristicIndex` in the other platforms so should we not do the same here?

https://github.com/chipweinberger/flutter_blue_plus/blob/5b70c11ae5a8fe3b797c8ed0eef2bcd6b4138b79/packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart#L550

https://github.com/chipweinberger/flutter_blue_plus/blob/5b70c11ae5a8fe3b797c8ed0eef2bcd6b4138b79/packages/flutter_blue_plus_web/lib/flutter_blue_plus_web.dart#L614

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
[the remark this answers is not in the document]
tnc1997: > It is not returned in the returned data structures.

If it is unused and not returned then should it be removed from the `BmDescriptorData` model?

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
tnc1997: > It is not returned in the returned data structures.

If it is unused and not returned then should it be removed from the `BmDescriptorData` model?

packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
tnc1997: Apologies if I am missing something but it looks like `false` is still returned in all cases below the try/catch?

mgeilich:
I am going to close this pull request.  I believe you want a solution that fully implements the Bluetooth spec for this feature.  I have implemented only the subset that I need for my own work.

At least for the time being, I will customize a private fork of FBP to support my work.  I hope that you can prioritize this feature in the future so that I can return to the public version of FBP.

PR #1080, opened 2024-12-16
distinguish between read and notification for future completion

hgrf:
Hi, I have run into an issue with flutter_blue_plus when I enable notifications for a characteristic and also call read operations. You might say "there is no reason to use read operations if you are using notifications" and I would accept that answer.

However, I did some digging and saw that the completion criteria for the future in https://github.com/chipweinberger/flutter_blue_plus/blob/c195c47989b3df6f9b514d7a82bcac158cf54ce5/lib/src/bluetooth_characteristic.dart#L120 does not distinguish between the result of a read operation or a notification. Now for Android, this distinction is pretty simple to introduce, c.f. this PR. For iOS / macOS unfortunately I do not know how to do it, maybe you have a suggestion?

There are two problems that might happen when not doing this distinction:

* the read value might correspond to a notification value (harmless in all cases where the values are "stable", i.e. there is no funky distinction between read and notify logic on the other side)
* a subsequent operation might fail with a BUSY error, because the actual read has not completed, but its future was completed by the notification

For some context, I ran into this problem when doing subsequent calls to setNotifyValue() and read(), here is the corresponding debug log:

```
I/flutter (13262): BLE: INFO: 2024-12-16 13:36:42.321029: reading characteristic 6d1a6389-697f-657e-0000-000000000000...
D/[FBP-Android](13262): [FBP] onMethodCall: readCharacteristic
D/[FBP-Android](13262): [FBP] onCharacteristicRead:
D/[FBP-Android](13262): [FBP]   chr: 6d1a6389-697f-657e-0000-000000000000
D/[FBP-Android](13262): [FBP]   status: GATT_SUCCESS (0)
I/flutter (13262): BLE: INFO: 2024-12-16 13:36:42.365419: reading characteristic 6d166384-6985-6544-0000-000000000000...
D/[FBP-Android](13262): [FBP] onMethodCall: readCharacteristic
D/[FBP-Android](13262): [FBP] onCharacteristicRead:
D/[FBP-Android](13262): [FBP]   chr: 6d166384-6985-6544-0000-000000000000
D/[FBP-Android](13262): [FBP]   status: GATT_SUCCESS (0)
I/flutter (13262): BLE: INFO: 2024-12-16 13:36:42.397289: activating notification for 6408736b-6169-637b-0000-000000000000...
D/[FBP-Android](13262): [FBP] onMethodCall: setNotifyValue
[...]

PR #1118, opened 2025-01-30
[Android] distinguish read and notification for future completion

hgrf:
Hi Chip,

sorry I did not reply to your comments on https://github.com/chipweinberger/flutter_blue_plus/pull/1080.

I have rebased it, rewritten it to use an enumeration and also made "receiveEvent" optional in the constructor of BmCharacteristicData, so that it does not break the Linux and web implementations. I think both of them support this distinction, so if you want me to, I can implement and test it.

### DIFF
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -459,6 +459,12 @@ class BmReadCharacteristicRequest {
   }
 }
 
+enum BmReceiveEventEnum {
+  unknown, // 0
+  read, // 1
+  notify, // 2
+}
+
 class BmCharacteristicData {
   final DeviceIdentifier remoteId;
   final Guid serviceUuid;
@@ -468,6 +474,7 @@ class BmCharacteristicData {
   final bool success;
   final int errorCode;
   final String errorString;
+  final BmReceiveEventEnum receiveEvent;
 
 
   BmCharacteristicData({
@@ -479,6 +486,7 @@ class BmCharacteristicData {
     required this.success,
     required this.errorCode,
     required this.errorString,
+    this.receiveEvent = BmReceiveEventEnum.unknown,
   });
 
   factory BmCharacteristicData.fromMap(Map<dynamic, dynamic> json) {
@@ -491,6 +499,7 @@ class BmCharacteristicData {
       success: json['success'] != 0,
       errorCode: json['error_code'],
       errorString: json['error_string'],
+      receiveEvent: BmReceiveEventEnum.values[json['receive_event']],
     );
   }
 }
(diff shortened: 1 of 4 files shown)


chipweinberger:
thanks for the update.

1. now that i see it in use, `receiveType` is a better name
2. will this compile and web & linux?

chipweinberger:
`.where((p) => p.receiveEvent != BmReceiveEventEnum.notify);`

don't like this so much.

ideally make it clear it is android only, or check for `indicate` or `unknown`

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -479,6 +486,7 @@ class BmCharacteristicData {
     required this.success,
     required this.errorCode,
     required this.errorString,
+    this.receiveEvent = BmReceiveEventEnum.unknown,
```
chipweinberger, line 489: lets make it required

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -468,6 +474,7 @@ class BmCharacteristicData {
   final bool success;
   final int errorCode;
   final String errorString;
+  final BmReceiveEventEnum receiveEvent;
```
chipweinberger, line 477: lets make this nullable

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -491,6 +499,7 @@ class BmCharacteristicData {
       success: json['success'] != 0,
       errorCode: json['error_code'],
       errorString: json['error_string'],
+      receiveEvent: BmReceiveEventEnum.values[json['receive_event']],
```
chipweinberger, line 502: lets make an explicit 'parse' helper function

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -459,6 +459,12 @@ class BmReadCharacteristicRequest {
   }
 }
 
+enum BmReceiveEventEnum {
+  unknown, // 0
+  read, // 1
+  notify, // 2
+}
```
chipweinberger, line 466: lets just have this:

```
enum BmReceiveEventEnum {
  read, // 1
  notify, // 2
}
```

and then make it nullable, and have a `BmReceiveType? _parseReceiveTypeFromInteger` function

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
tnc1997: Leaving it optional means that we need not make this a breaking change and would be better if possible.

We could default to `BmReceiveEventEnum.unknown` or we could make the field nullable as done for `pin`.

https://github.com/chipweinberger/flutter_blue_plus/pull/1119#discussion_r1937969385

chipweinberger: breaking is fine tho. I still dont really understand why we wouldn't want to break.

tnc1997: https://github.com/chipweinberger/flutter_blue_plus/pull/1119#discussion_r1938004017

PR #1119, opened 2025-01-30
[Android] add option to provide pairing PIN to createBond

hgrf:
Hi again,

I use flutter_blue_plus to communicate with GATT peripherals that have individual PINs. I can query the PINs from a Cloud database and show them in my app, but it would be great to enter the pairing PIN automatically if the platform allows it, which is the case on Android, not on iOS (not sure about bluez, but it probably is, even though createBond is Android-only for now).

### DIFF
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -1,3 +1,5 @@
+import 'dart:typed_data';
+
 import 'device_identifier.dart';
 import 'guid.dart';
 import 'log_level.dart';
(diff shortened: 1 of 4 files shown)


chipweinberger:
looks good to me

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -877,10 +879,19 @@ class BmBondStateResponse {
 
 class BmCreateBondRequest {
   DeviceIdentifier remoteId;
+  Uint8List? pin;
 
   BmCreateBondRequest({
     required this.remoteId,
+    required this.pin,
```
tnc1997, line 886: @chipweinberger could we remove the `required` here if possible so that this is not a breaking change?

```suggestion
    this.pin,
```

chipweinberger: its fine to break imo. The windows implementation is not started yet anyway.

tnc1997: > whats the motivation for not breaking?

Not breaking the interface keeps it compatible with all the implementations without needing any changes.

> its fine to break imo.

No problem but ensure that you bump the major version when publishing the platform interface package.

PR #1124, opened 2025-02-15
[Web] Add support for optional services

kubatatami:
Hey, 

In this PR I added support for [optional services](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice#optionalservices). 
It's a required option to use device if you set no filters or you just filter by name.
More details you can find [here](https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/requestDevice#acceptalldevices) and [here](https://webbluetoothcg.github.io/web-bluetooth/#dom-requestdeviceoptions-optionalservices)

### DIFF
--- a/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
+++ b/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
@@ -219,6 +219,7 @@ class FlutterBluePlus {
     bool androidLegacy = false,
     AndroidScanMode androidScanMode = AndroidScanMode.lowLatency,
     bool androidUsesFineLocation = false,
+    List<Guid> webOptionalServices = const [],
   }) async {
     // check args
     assert(removeIfGone == null || continuousUpdates, "removeIfGone requires continuousUpdates");
@@ -263,7 +264,8 @@ class FlutterBluePlus {
           continuousDivisor: continuousDivisor,
           androidLegacy: androidLegacy,
           androidScanMode: androidScanMode.value,
-          androidUsesFineLocation: androidUsesFineLocation);
+          androidUsesFineLocation: androidUsesFineLocation,
+          webOptionalServices: webOptionalServices);
 
       Stream<BmScanResponse> responseStream = FlutterBluePlusPlatform.instance.onScanResponse;
 
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -80,6 +80,7 @@ class BmScanSettings {
   final bool androidLegacy;
   final int androidScanMode;
   final bool androidUsesFineLocation;
+  final List<Guid> webOptionalServices;
 
   BmScanSettings({
     required this.withServices,
@@ -93,6 +94,7 @@ class BmScanSettings {
     required this.androidLegacy,
     required this.androidScanMode,
     required this.androidUsesFineLocation,
+    required this.webOptionalServices,
   });
 
   Map<dynamic, dynamic> toMap() {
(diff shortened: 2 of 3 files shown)


packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
```diff
@@ -219,6 +219,7 @@ class FlutterBluePlus {
     bool androidLegacy = false,
     AndroidScanMode androidScanMode = AndroidScanMode.lowLatency,
     bool androidUsesFineLocation = false,
+    List<Guid> webOptionalServices = const [],
```
ekuleshov, line 222: @chipweinberger this somewhat looks odd being web-only parameter. Though there already Android-only params...

@kubatatami this parameter could also use some description how and why it is needed with the link to the docs you mentioned in the PR description

chipweinberger: @tnc1997 is the expert here. Yes the scan function is getting a little crowded. I think it's still fine though.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -80,6 +80,7 @@ class BmScanSettings {
   final bool androidLegacy;
   final int androidScanMode;
   final bool androidUsesFineLocation;
+  final List<Guid> webOptionalServices;
```
tnc1997, line 83: I would suggest making this nullable (or default to an empty list) to cover other non-web platforms.

```suggestion
  final List<Guid>? webOptionalServices;
```

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -93,6 +94,7 @@ class BmScanSettings {
     required this.androidLegacy,
     required this.androidScanMode,
     required this.androidUsesFineLocation,
+    required this.webOptionalServices,
```
tnc1997, line 97: I would suggest making this optional (or default to an empty list) to cover other non-web platforms.

```suggestion
    this.webOptionalServices,
```

packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
tnc1997: I agree that `webOptionalServices` makes sense here and follows the existing convention for other non-web platforms.

tnc1997 (approved):

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
chipweinberger: i disagree. just use an empty list

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
chipweinberger: I like using an empty list

chipweinberger:
looks good to me.

kubatatami:
Wow. I didn't expect you to accept it so quickly. Thanks!

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
tnc1997: > just use an empty list

```suggestion
this.webOptionalServices = const [],
```

> maybe I don't fully understand your motivation

I am simply attempting to reduce the amount of maintenance work required in each federated platform implementation as that is something that you originally raised concerns about.

chipweinberger: got it. so far its been pretty low key. kinda quiet even.

ekuleshov:
> looks good to me. will merge.

@chipweinberger I added api doc for the new scan parameter and bumped versions in pubspecs...

PR #1129, opened 2025-02-21
[Refactor] Specify gradle plugins declaratively in example.

panmari:
As opposed to declaratively. This is necessary to run with new java/gradle versions. See https://docs.flutter.dev/release/breaking-changes/flutter-gradle-plugin-apply. Jointly bumping version numbers for gradle and related for easier in a new flutter installation.

Tested by successfully compiling/installing example app on Android.


chipweinberger:
you can just delete all the files and run this

```
cd ./example                                                                  
flutter config --enable-android 
flutter create .
flutter run
```

this way we are up to date with the latest recommendation

panmari:
Done, the biggest difference was the additional move to kotlin.

chipweinberger:
you have extra linux changes here

chipweinberger:
Also, perhaps we should just delete all these files and change the docs to look like this:

Screenshot 2025-02-22 at 12 54 10 PM

ekuleshov:
> Also, perhaps we should just delete all these files and change the docs ...

@chipweinberger but what about Android manifest settings for bluetooth in the example app?

panmari:
Rebased and removed linux changes. I feel it's easier to keep the boilerplate code to allow easily running the example. Outside of big moves, like the gradle declaritive style I point out in the description, updates won't be needed.

chipweinberger:
> > Also, perhaps we should just delete all these files and change the docs ...
> 
> @chipweinberger but what about Android manifest settings for bluetooth in the example app?

good point. maybe we could just leave those files.

panmari:
Either solution seems fine and an improvement to the status quo. Do you want me to change the PR to doc-approached instead?

One minor consideration: If we expect folks to regenerate files that are not checked in to version control, the working directory will always be slightly polluted for contributors.

chipweinberger:
> maybe we could just leave those files.

we just need to test if this will work.

> the working directory will always be slightly polluted for contributors.

we can use gitignore

chipweinberger:
So I tried the doc, approach  and it works. 

But we also need to change minSdkVersion from `minSdkVersion flutter.minSdkVersion` to `minSdkVersion 21`, so it requires 2 changes.

Issue #1059, opened 2024-11-22
[Perf]: use bytes instead of hex for platform communication


chipweinberger:
### What is your feature request?

In FBP, we we use hex strings for platform communication 

**Android**: `hexToBytes`, `bytesToHex`
**iOS**: `convertHexToData`,`convertDataToHex`

We should instead switch to raw bytes, and convert to hex only when we log.

chipweinberger:
Anyone can implement this and open a PR if they want.

chipweinberger:
PR: https://github.com/chipweinberger/flutter_blue_plus/pull/1067

chipweinberger:
@tnc1997 , this would also be nice to improve in the interface. But not super critical.

PR #1130, opened 2025-02-23
use bytes instead of hex for platform communication #1059

ekuleshov:
Recreated an old PR for #1059


tnc1997 (approved):
These changes look good to me, don't forget to bump the _major_ version of the platform interface, as this is a breaking change.

chipweinberger:
awesome. thanks!!

chipweinberger (approved):

PR #1137, opened 2025-03-01
Assert web optional services is only set for web.

panmari:
The argument won't have an effect otherwise. Specifying it currently on android currently causes an obscure failure.


chipweinberger:
it's already clear from the name it's just for web.

PR #1138, opened 2025-03-01
web: More gracefully handle errors from getCharacteristics()

panmari:
On failure, return a service without characteristics. I ended up in this state when specifying during scan a webOptionalServices that didn't exist on the device.


chipweinberger:
I don't think we want this.

PR #1142, opened 2025-03-03
fix: Correct casting of raw json in BmScanAdvertisement

panmari:
Currently fails with

I/flutter ( 3579): type '_Map<dynamic, dynamic>' is not a subtype of type 'Iterable<MapEntry<int, List<int>>>'

Surprisingly, the exception is swallowed somewhere without producing logs. I found this with some trial and error.

I tried some iterations of a fix and this is the most elegant way of making it work. Tested with example app on Android. Without the fix, devices don't show up, as they don't make it to dart. Surprisingly, the exception is swallowed somewhere without producing logs.


chipweinberger:
seems okay to me.

@ekuleshov thoughts? you changed this code here https://github.com/chipweinberger/flutter_blue_plus/commit/b9e1a0d6f271d2537e9e9c598dae53947b6d9f89

PR #1143, opened 2025-03-04
Fix: conversion code for devices with service data and uuids.

panmari:
The raw key/elements are string here and need to be converted to Guid.

### DIFF
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -158,11 +158,11 @@ class BmScanAdvertisement {
     // Cast the data to the right type
     Map<Guid, List<int>> serviceData = {};
     rawServiceData.forEach((k, v) {
-      serviceData[k] = v;
+      serviceData[Guid(k)] = v;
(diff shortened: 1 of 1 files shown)


chipweinberger:
isn't this the code you just changed?

panmari:
I wasn't able to test it with all types of devices at first, sorry about that. Is it feasible to add unit test coverage for this part of code?

ekuleshov:
> I wasn't able to test it with all types of devices at first, sorry about that. Is it feasible to add unit test coverage for this part of code?

We should be able to use a "mock" platform code that could return some pre-populated response messages from the platform channels.

@chipweinberger are you open to dragging in [mockito](https://pub.dev/packages/mockito) as a dev dependency? we would need to generate some mocks with it. @tnc1997 what's your take on this?

tnc1997:
>  @tnc1997 what's your take on this?

I did previously write a number of unit tests that covered the platform interface work for improved confidence but these were unwanted and so they were excluded from the changes.

chipweinberger:
TBH I don't think we need tests for this project. Just don't think they add much value for this project.

chipweinberger:
Personally I don't think they're worth the time in this project.

PR #1162, opened 2025-03-29
Fix exception when adapters is empty.

panmari:
On some linux platforms, a disabled adapter is represented as missing.

Currently throws `Unhandled Exception: Bad state: No element`


chipweinberger:
thanks

PR #1169, opened 2025-04-09
better Guid equals operator

ekuleshov:
A small fix for Guid to improve equals method contract to handle hashcode collisions in maps.


chipweinberger:
good catch. that's a bug for sure.

Issue #1167, opened 2025-04-06
[Feature]: consider throwing an error from scan on Android when location services are not available/disabled


ekuleshov:
### FlutterBluePlus Version

1.35.3

### Flutter Version

3.29.2

### What OS?

Android

### OS Version

9+

### Bluetooth Module

n/a

### What is your feature request?

Doing a BLE scan on Android device with location services disabled silently returns no devices found.
In order to inform users that location services are required it would be helpful to throw an error when location services are disabled or not available on a given device.

If that make sense I can work on a PR for this.

### Logs

```shell
n/a
```

chipweinberger:
i agree. a pr would be great.

chipweinberger:
merged. 1.35.4

PR #1170, opened 2025-04-09
[Android] check if location services are enabled when doing scan (#1167)

ekuleshov:
This implements #1167 to throw error on scan when Android location services are disabled.
I also updated a few permission-related error messages.

Tested on Android 14, 9 and 7 using the example app.


chipweinberger:
LGTM

tnc1997 (approved):

PR #1175, opened 2025-04-16
Fix errorCode for createBond on iOS

kevin49999:
Seems to accidentally have been set to another method `setPreferredPhy`


ekuleshov:
LGTM

PR #1183, opened 2025-04-20
Fix: handle null manufData and avoid variable redefinition

ishaq2321:
Removed duplicate definition of manufDataB HashMap variable
Replaced references to undefined manufData variable with rawMsd (which was correctly obtained from the scan record)
Fixed the data processing logic to properly extract manufacturer data from raw bytes
Corrected the conditional check for adding manufacturer data to the result map


ekuleshov:
Thank you for the PR. However why there is a change in permission handling?

Also, was ot necessary to move manufacturer data handling code around?

Those things and formatting changes are making it harder to review, verify and accept this pull request.

PR #1184, opened 2025-04-21
fix compilation errors on Android

ekuleshov:
@chipweinberger @tnc1997 a rough fix for the Android compilation errors. These are  needed for an urgent release of the Android and the root federated plugin.

I'd also like to remove duplicated null check at the end of the Android's `bmScanAdvertisement()` method and put things into the returned map as they processed in that method. That would also help with some unnecessary maps creation. Not sure if there other reasons not to do that.


chipweinberger:
should be split into different comits

ekuleshov:
> should be split into different commits, like this:
> 
> https://github.com/chipweinberger/flutter_blue_plus/commits/master/
> 
> also, you can do releases just by pushing tags. see master branch

@chipweinberger I'm on the road for a couple more days and somewhat fuzzy about the release process. Never done a release yet... 🙁

Could you cherry pick the individual ones from my branch and do a release?

chipweinberger:
ya ive been pretty busy too but no worries ill try to get to it.

thanks for the pr

chipweinberger:
fixed now. too lazy to bump the main package tho

chipweinberger:
pushed 4.0.2

PR #1185, opened 2025-04-21
fix build deprecation warnings on Android

ekuleshov:
This should fix Java 8 obsolete build warning below. The Flutter's core plugins been on Java 11 target for a while.

Verified using `flutter build appbundle` with the example app.

```
Running Gradle task 'bundleRelease'...                          
warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.
3 warnings
```


chipweinberger:
great

PR #1191, opened 2025-04-25
[Refactor] Return empty name if adaptors list empty

tnc1997:


packages/flutter_blue_plus_linux/lib/flutter_blue_plus_linux.dart
```diff
@@ -438,7 +438,9 @@ final class FlutterBluePlusLinux extends FlutterBluePlusPlatform {
     await _initFlutterBluePlus();
 
     return BmBluetoothAdapterName(
-      adapterName: _client.adapters.first.name,
+      adapterName: _client.adapters.length > 0
+          ? _client.adapters.first.name
+          : '',
```
ekuleshov, line 443: @tnc1997 you can use `_client.adapters.firstOrNull?.name ?? '',` 

There are few other places where `_client.adapters.first` is being used. Perhaps they should also be changed to use `firstOrNull`.

BTW, the Dart lints say better to use `isEmpty` instead of the `.length` check.

Issue #1192, opened 2025-04-25
[Help]: Location Services should not be required for Bluetooth scan


keval-devani:
### Requirements

- [x] I've looked at the README 'Common Problems' section

### Have you checked this problem on the example app?

Yes

### FlutterBluePlus Version

1.35.4

### Flutter Version

3.29.3

### What OS?

macOS

### OS Version

15.3.2

### Bluetooth Module

Scanning for BLE device

### What is your problem?

It looks like we misplaced the isLocationEnabled check. If the app doesn't intend to use location for scanning, there's no need to check for it at all. This unnecessary check was introduced in https://github.com/chipweinberger/flutter_blue_plus/issues/1167 [code](https://github.com/chipweinberger/flutter_blue_plus/blob/master/packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java#L504)

According to the [FlutterBluePlus documentation](https://pub.dev/packages/flutter_blue_plus#add-permissions-for-android-no-location), location permissions (and checks like isLocationEnabled) are only required if the app actually uses location-based scanning

### Logs

```shell
PlatformException(startScan, Location services are required for Bluetooth scan, null, null)
```

ekuleshov:
@keval-devani this is not a location permission check but a check for system location services.

I can't check on Android 15 right now, but on 10..12 the location services are required even when bluetooth scan permission. Even if you give permission the BLE scan returns nothing with location services disabled.

keval-devani:
@ekuleshov I tested version `flutter_blue_plus: 1.35.3` Android was scanning and returning results on android 14 and now on android 15 that I can verify without **location service turn on**. 

Please see here https://developer.android.com/develop/connectivity/bluetooth/bt-permissions

keval-devani:
@ekuleshov Yes I agree with your response on article. 

Right now I'm checking on Android 9 (API Level 28) using `flutter_blue_plus: 1.35.3` with location service is turn off and getting scanResults 🤨

William-Geng:
Can confirm! when using 1.35.4 with flutter_blue_plus_android 4.0.2, on android 14 and 15 you would get this error if you try to start scan with location service off.

Downgrading to an older version like I.34.5 would solve the issue - did anything change?

chipweinberger:
**Location Services** are required for scanning. **Location Permissions** are _not_ required.

chipweinberger:
closing.

[...]

PR #1199, opened 2025-04-26
[Android] location services check

ekuleshov:
This should address #1192 issue.

@chipweinberger @tnc1997 what are the steps to release these changes?

Need to bump version in the platform interface. With a new attribute added should it be `4.0.1 -> 4.0.2` or we need a higher version?

Similarly for Android plugin. Is `4.0.2 -> 4.0.3` ok for a new property? And the Android plugin probably shouldn't depend on `^4.0.0` version of platform as it needs newly added properties...

### DIFF
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -80,6 +80,7 @@ class BmScanSettings {
   final bool androidLegacy;
   final int androidScanMode;
   final bool androidUsesFineLocation;
+  final bool androidCheckLocationServices;
   final List<Guid> webOptionalServices;
 
   BmScanSettings({
@@ -94,6 +95,7 @@ class BmScanSettings {
     required this.androidLegacy,
     required this.androidScanMode,
     required this.androidUsesFineLocation,
+    required this.androidCheckLocationServices,
     required this.webOptionalServices,
   });
 
@@ -110,6 +112,7 @@ class BmScanSettings {
     data['android_legacy'] = androidLegacy;
     data['android_scan_mode'] = androidScanMode;
     data['android_uses_fine_location'] = androidUsesFineLocation;
+    data['android_check_location_services'] = androidCheckLocationServices;
     data['web_optional_services'] = webOptionalServices.map((s) => s.str).toList();;
     return data;
   }
--- a/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
+++ b/packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
@@ -95,7 +95,7 @@ class BmScanSettings {
     required this.androidLegacy,
     required this.androidScanMode,
     required this.androidUsesFineLocation,
-    required this.androidCheckLocationServices,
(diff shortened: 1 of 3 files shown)


tnc1997:
> With a new attribute added should it be `4.0.1 -> 4.0.2` or we need a higher version?

The new attribute is `required` in the constructor and therefore it is a [breaking change](https://stackoverflow.com/a/72089688) so it should be 5.0.0.

If the new attribute can have a [default](https://github.com/chipweinberger/flutter_blue_plus/blob/ee1258bbb58b03732981bdfd8b7bea7e9a9d1f54/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart#L230-L233) value in order to not make it `required` then the version can be 4.0.2.

```dart
  BmScanSettings({
    required this.withServices,
    required this.withRemoteIds,
    required this.withNames,
    required this.withKeywords,
    required this.withMsd,
    required this.withServiceData,
    required this.continuousUpdates,
    required this.continuousDivisor,
    this.androidLegacy = false,
    this.androidScanMode = 2,
    this.androidUsesFineLocation = false,
    this.androidCheckLocationServices = true,
    required this.webOptionalServices,
  });
```

tnc1997:
> Similarly for Android plugin. Is 4.0.2 -> 4.0.3 ok for a new property?

As long as the API exposed by the Android plugin does not have breaking changes then that version bump is perfect.

> And the Android plugin probably shouldn't depend on `^4.0.0` version of platform as it needs newly added properties

Android will need its platform interface constraint increased to `^5.0.0` or `^4.0.2` depending on the breaking change.

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
```diff
@@ -94,6 +95,7 @@ class BmScanSettings {
     required this.androidLegacy,
     required this.androidScanMode,
     required this.androidUsesFineLocation,
+    required this.androidCheckLocationServices,
```
tnc1997, line 98: Could we give this new attribute a sensible default so that it will not be a breaking change to the interface?

```suggestion
    this.androidCheckLocationServices = true,
```

tnc1997 (approved):

ekuleshov:
> If the new attribute can have a [default](https://github.com/chipweinberger/flutter_blue_plus/blob/ee1258bbb58b03732981bdfd8b7bea7e9a9d1f54/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart#L230-L233) value in order to not make it `required` then the version can be 4.0.2.

Right. For some reason FBP API uses defaults, but all msg classes have all fields `required`. Personally I'd prefer the defaults... @chipweinberger what's your take?

Also, when versions are settled. Should I push the `CHANGELOG.md` and `pubspec.yaml` changes for platform interface and tag it to release?

Then push Android plugin changes?

packages/flutter_blue_plus_platform_interface/lib/src/bluetooth_msgs.dart
ekuleshov: @chipweinberger are you okay with this?

chipweinberger: sure

chipweinberger:
👍🏻

PR #1200, opened 2025-04-27
Flutter lints fixes

ekuleshov:
This PR fixes most of the linter issues.

Things that hasn't fixed:

* info: Don't invoke 'print' in production code
  _seem to be part of a custom logging and we need to either disable lint or add ignore everywhere_

* [info: Don't return 'null' from a function with a return type of 'void'. (avoid_returning_null_for_void at [flutter_blue_plus] lib/src/bluetooth_device.dart:371)](https://github.com/chipweinberger/flutter_blue_plus/blob/38a09f22ef1ae8789a398a227f0fce7a2369a547/packages/flutter_blue_plus/lib/src/bluetooth_device.dart#L368-L372)
  _not really sure how to fix this_

* [info: Don't return 'null' from a method with a return type of 'void'. (avoid_returning_null_for_void at [flutter_blue_plus] lib/src/flutter_blue_plus.dart:634)](https://github.com/chipweinberger/flutter_blue_plus/blob/38a09f22ef1ae8789a398a227f0fce7a2369a547/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart#L633-L634)
  _not sure how to fix this, but maybe it is about time to remove this one_

* [info: The null check operator shouldn't be used on a variable whose type is a potentially nullable type parameter. (null_check_on_nullable_type_parameter at [flutter_blue_plus_example] lib/utils/utils.dart:14)](https://github.com/chipweinberger/flutter_blue_plus/blob/5934d505f4625e9d17c7cd2171f684d07876516f/packages/flutter_blue_plus/example/lib/utils/utils.dart#L14)
  _the proposed fix is to replace `_latestValue!` with `_latestValue as T` but I'm not sure how to test this code in the example app_

### DIFF
--- a/packages/flutter_blue_plus/analysis_options.yaml
+++ b/packages/flutter_blue_plus/analysis_options.yaml
@@ -21,7 +21,7 @@ linter:
   # `// ignore_for_file: name_of_lint` syntax on the line or in the file
   # producing the lint.
   rules:
-    # avoid_print: false  # Uncomment to disable the `avoid_print` rule
+    avoid_print: false  # Uncomment to disable the `avoid_print` rule
     # prefer_single_quotes: true  # Uncomment to enable the `prefer_single_quotes` rule
 
 # Additional information about this file can be found at
--- a/packages/flutter_blue_plus/example/analysis_options.yaml
+++ b/packages/flutter_blue_plus/example/analysis_options.yaml
@@ -21,7 +21,7 @@ linter:
   # `// ignore_for_file: name_of_lint` syntax on the line or in the file
   # producing the lint.
   rules:
-    # avoid_print: false  # Uncomment to disable the `avoid_print` rule
+    avoid_print: false  # Uncomment to disable the `avoid_print` rule
     # prefer_single_quotes: true  # Uncomment to enable the `prefer_single_quotes` rule
 
 # Additional information about this file can be found at
--- a/packages/flutter_blue_plus/example/ios/Podfile.lock
+++ b/packages/flutter_blue_plus/example/ios/Podfile.lock
@@ -16,7 +16,7 @@ EXTERNAL SOURCES:
 
 SPEC CHECKSUMS:
   Flutter: e0871f40cf51350855a761d2e70bf5af5b9b5de7
-  flutter_blue_plus_darwin: 3ea4ec9133b377febcc8a70b28cd2d2dc9242bd9
+  flutter_blue_plus_darwin: 904ed52825285a15f60e9d90312337a09f11a785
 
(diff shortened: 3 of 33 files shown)


tnc1997 (approved):

tnc1997:
> * info: Don't invoke 'print' in production code
>   _seem to be part of a custom logging and we need to either disable lint or add ignore everywhere_

Could this potentially be an opportunity to use the Dart/Flutter [logging](https://pub.dev/packages/logging) package instead of a custom implementation?

chipweinberger:
lets please not add more dependencies that we dont control.

chipweinberger:
> not sure how to fix this, but maybe it is about time to remove this one

AI codegen still prefers using the original flutter blue.

ekuleshov:
> AI codegen still prefers using the original flutter blue.

Meaning?

> we dont need it.

Which one?

The one I wasn't sure about... Actually not sure of the meaning or intent of a `Stream<void>` in this code. Why not return `.map((m) => this)` ?

```
  Stream<void> get onServicesReset {
    return FlutterBluePlusPlatform.instance.onServicesReset
        .where((p) => p.remoteId == remoteId)
        .map((m) => null);
  }
```

chipweinberger:
```
Stream<void> get onServicesReset {
  return FlutterBluePlusPlatform.instance.onServicesReset
    .where((e) => e.remoteId == remoteId)
    .map<void>((_) { /* no return, so this is a void callback */ });
}
```

```
Stream<void> get onServicesReset {
  return FlutterBluePlusPlatform.instance.onServicesReset
    .where((e) => e.remoteId == remoteId)
    .map((_) => null)
    .cast<void>();
}
```

ekuleshov:
> > Meaning?
> there is still a lot of new code being written with `.instance`

Huh? It returns null right now.

ekuleshov:
@chipweinberger I pushed analyzer setting change to disable print lint, as well as fixes for 2 other linter issues. The only issue we have left is with the `.instance` returning null

chipweinberger:
just disable the `.instance`  lint. 

> Huh? It returns null right now.

1. ai writes code
2. user copy pastes it into IDE
3. the deprecation warning helps them fix it

ekuleshov:
> just disable the `.instance` lint.
> 
> > Huh? It returns null right now.
> 
> 1. ai writes code
> 2. user copy pastes it into IDE
> 3. the deprecation warning helps them fix it

Hmm... The code isn't just deprecated - it is broken there. Having a deprecation on it is kind of misleading and an odd argument about AI - it generates a broken code and the deprecation warning isn't really forcing you to fix it.

chipweinberger:
It's to help people know how to fix it.

PR #1201, opened 2025-04-28
[iOS] Fix crash on incorrect PIN

icooke-syna:
On iOS, reading an encrypted characteristic to trigger the PIN prompt, then entering an incorrect PIN causes a crash.

```
[FBP-iOS] didUpdateValueForCharacteristic:
[FBP-iOS]   chr: 318848ff-f881-44a8-9f65-25be4801594f
[FBP-iOS]   error: Encryption is insufficient.
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[4]'
```

Resolved by providing a null object for `value` in the result dictionary as is done for `primary_service_uuid`.


chipweinberger:
thanks

PR #1207, opened 2025-05-08
Support multiple characteristics with same uuid 

Moelfarri:
A first attempt at addressing issue #795  using getInstanceId() on android. The next step is to support descriptors of similar characteristics on android. The proposed solution aims to be backwards compatible. 

If you have any feedback or suggestions regarding what is the best approach for supporting descriptors on the android code then it is most welcome.

### DIFF
--- a/packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
+++ b/packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
@@ -863,6 +863,8 @@ public class FlutterBluePlusPlugin implements
                     String serviceUuid =        (String) data.get("service_uuid");
                     String characteristicUuid = (String) data.get("characteristic_uuid");
                     String primaryServiceUuid = (String) data.get("primary_service_uuid");
+                    Integer instanceId = (Integer) data.get("instance_id");
+
 
                     // check connection
                     BluetoothGatt gatt = mConnectedDevices.get(remoteId);
@@ -875,7 +877,7 @@ public class FlutterBluePlusPlugin implements
                     waitIfBonding();
 
                     // find characteristic
-                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid);
+                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid, instanceId);
                     if (found.error != null) {
                         result.error("readCharacteristic", found.error, null);
                         break;
@@ -912,6 +914,8 @@ public class FlutterBluePlusPlugin implements
                     byte[] value =              (byte[]) data.get("value");
                     int writeTypeInt =             (int) data.get("write_type");
                     boolean allowLongWrite =      ((int) data.get("allow_long_write")) != 0;
+                    Integer instanceId = (Integer) data.get("instance_id");
+
 
                     int writeType = writeTypeInt == 0 ?
                         BluetoothGattCharacteristic.WRITE_TYPE_DEFAULT :
@@ -928,7 +932,7 @@ public class FlutterBluePlusPlugin implements
                     waitIfBonding();
 
                     // find characteristic
-                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid);
+                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid, instanceId);
                     if (found.error != null) {
                         result.error("writeCharacteristic", found.error, null);
                         break;
@@ -1008,6 +1012,7 @@ public class FlutterBluePlusPlugin implements
                     String characteristicUuid = (String) data.get("characteristic_uuid");
                     String descriptorUuid =     (String) data.get("descriptor_uuid");
                     String primaryServiceUuid = (String) data.get("primary_service_uuid");
+                    Integer instanceId = (Integer) data.get("instance_id");
 
                     // check connection
                     BluetoothGatt gatt = mConnectedDevices.get(remoteId);
@@ -1020,13 +1025,14 @@ public class FlutterBluePlusPlugin implements
                     waitIfBonding();
 
                     // find characteristic
-                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid);
+                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid, instanceId);
                     if (found.error != null) {
                         result.error("readDescriptor", found.error, null);
                         break;
                     }
 
                     BluetoothGattCharacteristic characteristic = found.characteristic;
+
 
                     // find descriptor
                     BluetoothGattDescriptor descriptor = getDescriptorFromArray(descriptorUuid, characteristic.getDescriptors());
@@ -1056,6 +1062,7 @@ public class FlutterBluePlusPlugin implements
                     String descriptorUuid =     (String) data.get("descriptor_uuid");
                     String primaryServiceUuid = (String) data.get("primary_service_uuid");
                     byte[] value =              (byte[]) data.get("value");
+                    Integer instanceId = (Integer) data.get("instance_id");
 
                     // check connection
                     BluetoothGatt gatt = mConnectedDevices.get(remoteId);
@@ -1068,7 +1075,7 @@ public class FlutterBluePlusPlugin implements
                     waitIfBonding();
 
                     // find characteristic
-                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid);
+                    ChrFound found = locateCharacteristic(gatt, serviceUuid, characteristicUuid, primaryServiceUuid, instanceId);
                     if (found.error != null) {
                         result.error("writeDescriptor", found.error, null);
                         break;
@@ -1136,6 +1143,7 @@ public class FlutterBluePlusPlugin implements
                     String primaryServiceUuid = (String) data.get("primary_service_uuid");
                     boolean forceIndications =  (boolean) data.get("force_indications");
                     boolean enable =            (boolean) data.get("enable");
+                    Integer instanceId = (Integer) data.get("instance_id");
 
                     // check connection
                     BluetoothGatt gatt = mConnectedDevices.get(remoteId);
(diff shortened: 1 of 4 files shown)


chipweinberger:
nice. but we need to have a PR that supports all platforms.

Moelfarri:
Absolutely, I will try to patch this incrementally, so it is easy to review. 
I have added the necessary changes to support instance id on Android on the characteristic and descriptor level, and will try to add similar changes on the other platforms with one or two commits at a time so it is easy to follow

chipweinberger:
yes it's a tricky problem. 

Maybe should use a hash of it's descriptors

In all the BLE APIS that use duplicate CHR uuids, AFAIK they all *instead* use descriptors to identify them. So we can hash the data within the descriptor and use that as our identifier.

chipweinberger:
> Descriptors are optional metadata about the characteristic, what do we do in the case where we have two or more similar characteristics but with no descriptors?

unsupported. we can log an error or something

Moelfarri:
@chipweinberger  Now the proposed feature should be supported in all platforms. Let me know what you think once you have had time to review the PR so we can merge the changes.

chipweinberger:
its not a general solution.

Moelfarri:
How ever given the discussion on this i do like the idea of maintaining of our own mappings that gets populated during discovery instead of relying on hashes. 

i.e. this:
```
Map<ObjectReference, uniqueIncrementingInteger>
Map<uniqueIncrementingInteger, ObjectReference>
```



I was playing a little with the idea and on for example the web platform it could be solved something like this, which i also tested now briefly and it seemed to work: 


```dart
extension on BluetoothRemoteGATTCharacteristic {
  Future<int?> get instanceId async {
    return charToInstanceIdMap[this]?.id;
  }
}

class UniqueIncrementingInteger {
  static int _counter = 0;

  final int id;

  UniqueIncrementingInteger() : id = _counter++;
}

Map<UniqueIncrementingInteger, BluetoothRemoteGATTCharacteristic> instanceIdToCharMap = {};
Map<BluetoothRemoteGATTCharacteristic, UniqueIncrementingInteger> charToInstanceIdMap = {};
```

`UniqueIncrementingInteger` internal value gets incremented every time an instance of its is created and it is only created during discovery for each char object. 

In the for loop for chars during discovery, one can simply do this:

```dart
UniqueIncrementingInteger instanceId = UniqueIncrementingInteger();
instanceIdToCharMap[instanceId] = c;
charToInstanceIdMap[c] = instanceId;
```

chipweinberger:
awesome :) 

```dart
class UniqueIncrementingInteger {
  static int _counter = 0;

  static int next() {
    _counter++;
    return _counter;
  }
}
```

I'd probably define it this way. no need for fancy constructor stuff.

Moelfarri:
There should be different versions of this in the code but i solved it with adding `(instanceId == null || c.instanceId == instanceId) ` to the characteristic retrieval code for the different platforms/languages. Meaning that if instanceId is null then just use the UUID otherwise match on UUID & instanceId. 

´´´dart 
  for (final c in array) {
      if (c.uuid == uuid) {
        if (instanceId == null || c.instanceId == instanceId) {
          return c;
        }
      }
´´´

chipweinberger:
looks good

Moelfarri:
Let me know if there’s anything else needed from my side on this PR. Otherwise, it looks like everything is ready to be merged. Thanks again for the great feedback throughout!

chipweinberger:
wow impressive! which platforms have you been able to test so far?

chipweinberger:
@tnc1997 @ekuleshov

chipweinberger:
looking at the code i see one issue

resetInstanceIds should just reset the instanceIds of that device, not any others.

Moelfarri:
I currently have tested Android, iOS, MacOS and Web and can verify that the solution works well on these platforms. I tried to run the Linux on a VM but unfortunately it has some limitations.

chipweinberger:
you can access the device id from the characteristic 

at least on ios

chr.service.peripheral

and same for descriptors 

so we can just iterate through the map and delete the chrs from our device id

Moelfarri:
I currently do not have 2 BLEs to test the device id thing but i will look into this sometime next week. Also All descriptors are accessible through their Characteristics though? Would it make sense to explicitly point to descriptor references?

chipweinberger:
> Also All descriptors are accessible through their Characteristics though? Would it make sense to explicitly point to descriptor references?

Yes I guess you are right actually. I think there is no need, unless duplicate descriptors are possible. I'm not sure.

> I currently do not have 2 BLEs to test the device id thing but i will look into this sometime next week.

 it's fine to write the code and test with a single device.

Moelfarri:
@chipweinberger @tnc1997 @ekuleshov Would love to have your feedback so we can move forward with this feature 😊

chipweinberger:
im happy to merge it

packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
```diff
@@ -2402,9 +2415,10 @@ public void onCharacteristicWrite(BluetoothGatt gatt, BluetoothGattCharacteristi
             String serviceUuid = uuidStr(characteristic.getService().getUuid());
             String characteristicUuid = uuidStr(characteristic.getUuid());
             String primaryServiceUuid = primaryService != null ? uuidStr(primaryService.getUuid()) : "";
+            Integer instanceId = characteristic.getInstanceId();
 
             // what data did we write?
-            String key = remoteId + ":" + serviceUuid + ":" + characteristicUuid + ":" + primaryServiceUuid;
+            String key = remoteId + ":" + serviceUuid + ":" + characteristicUuid + ":" + primaryServiceUuid + ":" + instanceId;
             byte[] value = mWriteChr.get(key) != null ? mWriteChr.get(key) : new byte[0];
             mWriteChr.remove(key);
```
ekuleshov, line 2423: @chipweinberger @Moelfarri it seems like this `key` value potentially could be different from `key` used for `mWriteChr.put()` call - it uses `noinst` there which never going to be present on remove here because `instanceId` is never null here.

It may need to either use the "default" value for `instanceId` (returned from `characteristic.getInstanceId()`) in all places instead of `noinst` or replace the "default" value here with `noinst`

chipweinberger: wow good job noticing this!! 

can you push a commit to fix it? if you can.

ekuleshov: @chipweinberger I was just merging the main into the Android cleanup PR...

Unfortunately I don't think I'll be able to look at this in the next few weeks. Also not sure what is the default non-null value returned by `characteristic.getInstanceId()` and the [javadoc is not helping either](https://developer.android.com/reference/android/bluetooth/BluetoothGattCharacteristic#getInstanceId()). Gemini says it returns 0 by default. We probably should use it instead of `noinst` string everywhere, even on the Dart side of things.

Ideally I'd need to make up a device with multiple chars to test this.

chipweinberger:
@Moelfarri can you open a PR to address these issues? thanks

Moelfarri:
I can address it next week when I am back from my travels 😊

chipweinberger:
@Moelfarri Ive added you as a collaborator! welcome to the group :) 

@tnc1997 @ekuleshov

packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
chipweinberger: we ought to do these fixes, before we forget

chipweinberger:
@Moelfarri can you finish these changes/fixes?

packages/flutter_blue_plus_android/android/src/main/java/com/lib/flutter_blue_plus/FlutterBluePlusPlugin.java
chipweinberger: so im looking at this code, and underwhat circumstance would instanceId be null? 

only if the user is doing custom things and instantiating their own objects, i would think.

in other words maybe we should just make instanceId required

chipweinberger:
Just ran the code for the first time. There are also these warnings. Surprised you didn't notice them!

fixed here: b9a2309e06f2c7d8a217e5fe1c5450c31d4c32f2

```
/Volumes/User/MBP-Google-Drive/jamcorder/app/pubdev/flutter_blue_plus/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m:2286:38: warning: incompatible pointer types sending 'NSString *' to parameter of type 'NSValue * _Nonnull' [-Wincompatible-pointer-types]
 2286 |             if ([charToInstanceIdMap[key] isEqual:instanceId]) {
      |                                      ^~~
In module 'Foundation' imported from /Volumes/User/MBP-Google-Drive/jamcorder/app/pubdev/flutter_blue_plus/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m:6:
   46 | - (nullable ObjectType)objectForKeyedSubscript:(KeyType)key API_AVAILABLE(macos(10.8), ios(6.0), watchos(2.0), tvos(9.0));
      |                                                         ^
/Volumes/User/MBP-Google-Drive/jamcorder/app/pubdev/flutter_blue_plus/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m:2292:53: warning: incompatible pointer types sending 'NSString *' to parameter of type 'NSValue * _Nonnull' [-Wincompatible-pointer-types]
 2292 |             [charToInstanceIdMap removeObjectForKey:keyToRemove];
      |                                                     ^~~~~~~~~~~
In module 'Foundation' imported from /Volumes/User/MBP-Google-Drive/jamcorder/app/pubdev/flutter_blue_plus/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m:6:
  101 | - (void)removeObjectForKey:(KeyType)aKey;
      |                                     ^
2 warnings generated.
warning: Run script build phase 'Run Script' will be run during every build because it does not specify any outputs. To address this issue, either add output dependencies to the script phase, or configure it to run in every build by unchecking "Based on dependency analysis" in the script phase. (in target 'Flutter Assemble' from project 'Runner')
```

chipweinberger:
Also just noticing we have

```
const NSMutableDictionary<NSNumber *, CBCharacteristic *> *instanceIdToCharMap = nil;
NSMutableDictionary<NSValue *, NSNumber *> *charToInstanceIdMap = nil;
```

instead of 

```
const NSMutableDictionary<NSNumber *, CBCharacteristic *> *instanceIdToCharMap = nil;
NSMutableDictionary<CBCharacteristic *, NSNumber *> *charToInstanceIdMap = nil;
```

why would we be using NSValue *? I dont understand...

Issue #1202, opened 2025-04-29
[Feature]: iOS support CBConnectPeripheralOptionRequiresANCS and CBConnectPeripheralOptionEnableTransportBridgingKey


wtuyp:
### FlutterBluePlus Version

1.35.2

### Flutter Version

3.29.2

### What OS?

iOS

### OS Version

15.7

### Bluetooth Module

None

### What is your feature request?

 iOS connect peripheral with options doesn't support CBConnectPeripheralOptionRequiresANCS and CBConnectPeripheralOptionEnableTransportBridgingKey options.
Please add them to connect options. Thanks!

### Logs

```shell
Future<void> connect({
    Duration timeout = const Duration(seconds: 35),
    int? mtu = 512,
    bool autoConnect = false,
})

// options
NSMutableDictionary *options = [[NSMutableDictionary alloc] init];
if (@available(iOS 17, *)) {
    [options setObject:autoConnect forKey:@"kCBConnectOptionEnableAutoReconnect"];
}
[self.centralManager connectPeripheral:peripheral options:options];
```

chipweinberger:
just open a pr. we arent free labor :)

PR #1221, opened 2025-06-09
iOS add CBConnectPeripheralOptionRequiresANCS and CBConnectPeripheralOptionEnableTransportBridgingKey options when connect device.

wtuyp:
iOS add CBConnectPeripheralOptionRequiresANCS and CBConnectPeripheralOptionEnableTransportBridgingKey options when connect device.

### DIFF
--- a/packages/flutter_blue_plus/lib/src/bluetooth_device.dart
+++ b/packages/flutter_blue_plus/lib/src/bluetooth_device.dart
@@ -97,10 +97,14 @@ class BluetoothDevice {
   ///      - auto connect is turned off by calling `disconnect`
   ///      - auto connect results in a slower connection process compared to a direct connection
   ///        because it relies on the internal scheduling of background scans.
+  ///   [enableTransportBridging] iOS only. This option tells the system to connect non-GATT profiles on classic Bluetooth devices, if there is a low energy GATT connection to the same device.
+  ///   [requiresANCS] iOS only. An option to require Apple Notification Center Service (ANCS) when connecting a device..
   Future<void> connect({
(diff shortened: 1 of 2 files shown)


wtuyp:
#1202

packages/flutter_blue_plus/lib/src/bluetooth_device.dart
```diff
@@ -97,10 +97,14 @@ class BluetoothDevice {
   ///      - auto connect is turned off by calling `disconnect`
   ///      - auto connect results in a slower connection process compared to a direct connection
   ///        because it relies on the internal scheduling of background scans.
+  ///   [enableTransportBridging] iOS only. This option tells the system to connect non-GATT profiles on classic Bluetooth devices, if there is a low energy GATT connection to the same device.
+  ///   [requiresANCS] iOS only. An option to require Apple Notification Center Service (ANCS) when connecting a device..
```
ekuleshov, line 101: @chipweinberger in other parts of the API we prefixed platform specific properties like `android*` (e.g. device scan).

Perhaps these new ios-specific properties should be also prefixed with "ios"

Also, while we are at it, what about other ios connection parameters?
https://developer.apple.com/documentation/corebluetooth/peripheral-connection-options?language=objc

chipweinberger:
yes please prefix with "ios"

packages/flutter_blue_plus/lib/src/bluetooth_device.dart
wtuyp: These options are commonly used on smartwatches.

PR #1229, opened 2025-07-12
move logging from FlutterBluePlus class to FlutterBluePlusPlatform

heyteacher:
I propose this changes in order to collect all logs from platform, moving logging functionality from  `FlutterBluePlus` to `FlutterBluePlusPlatform`.

In this way, listeners of `FlutterBluePlus.logs` stream really catch all logs yield by all platform packages (not only from package `FlutterBluePlus`).

I test only `android` platform.

### DIFF
--- a/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
+++ b/packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
@@ -635,7 +635,7 @@ class FlutterBluePlus {
   static Future<List<BluetoothDevice>> get connectedSystemDevices => systemDevices([Guid("1800")]);
 
   @Deprecated('No longer needed, remove this from your code')
-  static void get instance {}
+  static void get instance => null;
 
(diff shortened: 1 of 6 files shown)


packages/flutter_blue_plus/lib/src/flutter_blue_plus.dart
```diff
@@ -642,7 +635,7 @@ class FlutterBluePlus {
   static Future<List<BluetoothDevice>> get connectedSystemDevices => systemDevices([Guid("1800")]);
 
   @Deprecated('No longer needed, remove this from your code')
-  static void get instance => null;
+  static void get instance {}
```
chipweinberger, line 638: please undo this unrelated change

heyteacher: done. Sorry, linter marks as dartavoid_returning_null_for_void warning and I fix it automatically.

PR #1259, opened 2025-09-10
Wipe known services on device disconnection

devgianlu:
This bug has been observed on Android, but I think the fix applies to all platforms.

The following piece of code doesn't work:
```dart
BluetoothDevice btDevice = ...

// First connection
await btDevice.connect();
if (btDevice.servicesList.isEmpty) { // True
  await btDevice.discoverServices(); // Called, OK
}

print("${btDevice.servicesList}"); // Has services, OK

// Something happened, we need to disconnect
await btDevice.disconnect();

// After a while, reconnect to the device
await btDevice.connect();
if (btDevice.servicesList.isEmpty) { // False, still has services from before
  await btDevice.discoverServices(); // Not called 
}

final char = btDevice.servicesList.first.characteristics.first;
await char.read(); // Will throw "primary service not found"
```

The above happens because there's an inconsistency between the services cached in the Flutter side and those in the Android side. When `gatt.getServices()` is called in the Android code, no services are available because `gatt.discoverServices()` has not been called after the disconnection.

I think it's sane to wipe the known services once the device is disconnected as they become stale. I haven't investigated whether there is more stuff to wipe.

### DIFF
--- a/packages/flutter_blue_plus/lib/src/bluetooth_device.dart
+++ b/packages/flutter_blue_plus/lib/src/bluetooth_device.dart
@@ -35,7 +35,8 @@ class BluetoothDevice {
 
   /// Get services
   ///  - returns empty if discoverServices() has not been called
-  ///    or if your device does not have any services (rare)
+  ///    after the most recent connection, or if your device
+  ///    does not have any services (rare)
   List<BluetoothService> get servicesList {
     BmDiscoverServicesResult? result = FlutterBluePlus._knownServices[remoteId];
     if (result == null) {
@@ -235,6 +236,9 @@ class BluetoothDevice {
             .fbpEnsureAdapterIsOn("disconnect")
             .fbpTimeout(timeout, "disconnect");
       }
+
+      // Wipe known services to avoid inconsistent state between Flutter and the platform
+      FlutterBluePlus._knownServices.remove(remoteId);
 
       if (!kIsWeb && Platform.isAndroid) {
         // Disconnected, remove connect timestamp


chipweinberger:
you should always rediscover services after every connection. 

this would encourage bad code. Do you have another example?

chipweinberger:
okay, after looking closer at your example, yes this is an improvement

this is certainly a rather large breaking change though.

chipweinberger:
just realized you put this in the wrong place. it was supposed to go here. you're not handling autoconnect

```
      FlutterBluePlusPlatform.instance.onConnectionStateChanged.listen((r) {
        _connectionStates[r.remoteId] = r;
        if (r.connectionState == BmConnectionStateEnum.disconnected) {
          // clear mtu
          _mtuValues.remove(r.remoteId);

          // clear lastDescs (resets 'isNotifying')
          _lastDescs.remove(r.remoteId);

          // clear lastChrs (api consistency)
          _lastChrs.remove(r.remoteId);

          // cancel & delete subscriptions
          _deviceSubscriptions[r.remoteId]?.forEach((s) => s.cancel());
          _deviceSubscriptions.remove(r.remoteId);

          // Note: to make FBP easier to use, we do not clear `knownServices`,
          // otherwise `servicesList` would be more annoying to use. We also
          // do not clear `bondState`, for faster performance.

          // autoconnect
          if (!kIsWeb && Platform.isAndroid == false) {
            if (_autoConnect.contains(r.remoteId)) {
              if (_adapterStateNow == BmAdapterStateEnum.on) {
                var d = BluetoothDevice(remoteId: r.remoteId);
                d.connect(autoConnect: true, mtu: null).onError((e, s) {
                  if (logLevel != LogLevel.none) {
                    FlutterBluePlusPlatform.log("[FBP] [AutoConnect] connection failed: $e");
                  }
                });
              }
            }
          }
        }
      });
```

chipweinberger:
```
await char.read(); // Will throw "primary service not found"
```

are you sure? It should throw "device disconnected"

im going to undo this PR. There is no need for it and makes servicesList annoying to traverse.

devgianlu:
Yes I am sure, why would it throw disconnected? It's connected, but it doesn't know about the services.

chipweinberger:
oh okay, it's just that you have not called `discoverServices` yet?

you are supposed to call that after every connection. As per the documentation & readme.

chipweinberger:
that said, i agree this error message could be better.

it would be better if we returned `must call discoverServices`

feel free to open a PR to improve the error message.

chipweinberger:
^ the way i think it would work. Have both:

* _knownServices (never cleared)
* _discoveredChrs (cleared after disconnection)

If you call `await chr.read()`, it will check that the chr is in `_discoveredChrs`, so we can throw a "must call discoverServices" error instead of a "primary service not found" error.

Issue #1098, opened 2025-01-15
[Help]: Calling showPowerAlert = false does not prevent app from showing "Turn on Bluetooth to allow .." iOS popup every time


spehj:
### Requirements

- [X] I've looked at the README 'Common Problems' section

### Have you checked this problem on the example app?

No

### FlutterBluePlus Version

1.34.5

### Flutter Version

3.27.1

### What OS?

iOS

### OS Version

15.8.3 and higher

### Bluetooth Module

Don't know

### What is your problem?

Bluetooth feature is not mandatory for using our app. However, if we disable Bluetooth in iOS settings (or control panel) we get the popup "Your App would like to use Bluetooth for new connections. You can allow new connections in Settings.".

This happens every time opening the app without BLE turned on. It's annoying for our users.

We set the showPowerAlert option to false right at the start of the app in the Flutter main method, before calling any other FlutterBluePlus method:
```dart
Future<void> main() async {
  await FlutterBluePlus.setOptions(showPowerAlert: false);
// Other initializations

}
```

In the FBP documentation it's said that this needs to be set before any other method of the package:
```dart
 /// Set configurable options
  ///   - [showPowerAlert] Whether to show the power alert (iOS & MacOS only). i.e. CBCentralManagerOptionShowPowerAlertKey
  ///       To set this option you must call this method before any other method in this package.
  ///       See: https://developer.apple.com/documentation/corebluetooth/cbcentralmanageroptionshowpoweralertkey
  ///       This option has no effect on Android.
```

How can this system popup be prevented from being shown every time the app is started?

Example:
![anyway-to-stop-the-app-from-asking-for-bluetooth-access-v0-1vnx5nndtm1d1-ezgif com-webp-to-jpg-converter](https://github.com/user-attachments/assets/d94ee8af-abd6-4600-988f-ff82b937c031)

### Logs

```shell
No logs since I don't have a MAC right here at the moment.
```

chipweinberger:
needs logs

```
            Log(LDEBUG, @"showPowerAlert: %@", [self.showPowerAlert boolValue] ? @"yes" : @"no");
            Log(LDEBUG, @"restoreState: %@", [self.restoreState boolValue] ? @"yes" : @"no");
```

spehj:
@chipweinberger Here are the logs, when we restart the app (with `await FlutterBluePlus.setOptions(showPowerAlert: false);`
 in the Flutter main method):

[...]

PR #1268, opened 2025-09-16
[iOS] Initialise showPowerAlert to false

Ninban:
See https://github.com/chipweinberger/flutter_blue_plus/issues/1098, I found this issue late while investigating but it describes exactly the problem I was also having.

I tried to avoid introducing a breaking change, but I'm out of ideas.

# Description
`showPowerAlert`'s default value becomes `@(NO)`.

When a flutter application with FBP starts, `handleMethodCall` gets called with argument `flutterRestart` before anything in the `main` method can execute, including the call to FBP's `setOptions`.

This results in the CBCentralManager always initialising with the default values, implying that `CBCentralManagerOptionShowPowerAlertKey` starts at true. 

However, calling `setOptions` with `showPowerAlert = false` after the CBCentralManger was initialised does not apply the change, firstly because the previous implementation made it return immediately, so if it was already initialised changing the options did nothing. Then after testing, even if the CBCentralManager is reinitialised completely, the behaviour remains as if `showPowerAlert = true`.

Simply changing the default fixes the issue, but breaks functionality when `setOptions` is called with `showPowerAlert = true`. It behaves correctly again after reinitialising the CBCentralManager any time `setOptions` is called with different settings than what were previously used.

Strangely, I did another test that included the `CBCentralManagerOptionShowPowerAlertKey` key only if `showPowerAlert` was false, going by the [documented logic](https://developer.apple.com/documentation/corebluetooth/cbcentralmanageroptionshowpoweralertkey) that if the key wasn't included it defaulted to true anyway, yet the alert never shows up in that case in the first place.

## Testing done
Scenario 1:
Not calling `setOptions`: new default is false, therefore the power alert no longer shows. This is the breaking change.

Scenario 2:
Calling `setOptions` with `showPowerAlert = false`: the power alert does not show, as intended. Fixed by this PR.

Scenario 3:
Calling `setOptions` with `showPowerAlert = true`: the power alert shows up as intended. No changes here.

### DIFF
--- a/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
+++ b/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
@@ -48,6 +48,8 @@ typedef NS_ENUM(NSUInteger, LogLevel) {
 @property(nonatomic) LogLevel logLevel;
 @property(nonatomic) NSNumber *showPowerAlert;
 @property(nonatomic) NSNumber *restoreState;
+@property(nonatomic) NSNumber *initializedShowPowerAlert;
+@property(nonatomic) NSNumber *initializedRestoreState;
 @end
 
 @implementation FlutterBluePlusPlugin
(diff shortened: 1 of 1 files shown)


chipweinberger:
1. the default should be true
2. you need to call setOptions *before* any other FBP call. 
3. you should never reinitialize

this should probably be renamed to `setInitOptions` to make that make clear

btw, `flutterRestart` is only called *after* you call a FBP method

Ninban:
Thank you, I'll investigate more closely why `flutterRestart` gets called before I get the chance to call `setOptions` in my app (very first thing in main method after widgets binding) and submit a better fix if needed.

chipweinberger:
`setOptions` is not supposed to invoke `flutterRestart` for this reason.

it looks like this was a regression in 1.35.0

chipweinberger:
fixed: a2c5f05f4fc6b5a0218d9480360a4340333e9805

chipweinberger:
released 1.36.7

Ninban:
Confirming that did the trick. Thanks again!

PR #1318, opened 2026-02-11
fix(darwin): nil-guard rssi in didReadRSSI:error: to prevent crash

gotnull:
When CoreBluetooth calls peripheral:didReadRSSI:error: with a non-nil error (e.g. device disconnected mid-read), the rssi parameter can be `nil`. Inserting `nil` into an `NSDictionary` literal causes `NSInvalidArgumentException`, crashing the app.

This is a fatal crash, the app terminates immediately with no recovery path. It occurs when a BLE peripheral disconnects (spontaneously or intentionally) while an RSSI read is in flight.

The fix uses the nil-coalescing operator `(rssi ?: @(0))` to fall back to `0` when `rssi` is `nil`, consistent with how `error_string` and `error_code` are already guarded in the same dictionary.

Crash signature:
```
  NSInvalidArgumentException
  -[__NSPlaceholderDictionary initWithObjects:forKeys:count:]
  -[FlutterBluePlusPlugin peripheral:didReadRSSI:error:]
  FlutterBluePlusPlugin.m:1727
```


chipweinberger:
You are great. Thank you for finding this and fixing it.

chipweinberger:
released FBP 2.1.1

PR #1186, opened 2025-04-21
fixes Uint8List dereference on the iOS/macOS side (#1163)

ekuleshov:
This should fix #1163 
It seems like filtering by service uuids (and probably by manufacturer data) was also broken on ios/macos :(

PR #1334, opened 2026-06-04
fix(darwin): reject NSNull in setOptions to prevent crash on Flutter 3.10+

### LINKED ISSUE #1333: [Bug]: iOS/macOS crash on first BLE method call — '-[NSNull boolValue]: unrecognized selector' from setOptions
**Plugin version:** `flutter_blue_plus_darwin` 9.0.0 (via `flutter_blue_plus` 2.3.4)
**Flutter / Dart:** Flutter 3.41.x / Dart 3.11.x (also reproduced on 3.44.x / Dart 3.13.x)
**Platforms affected:** iOS (Release & Debug), macOS — Android unaffected

## What happens

App crashes immediately when calling **any** BLE method (e.g. `FlutterBluePlus.systemDevices`, `FlutterBluePlus.adapterState.first`, `FlutterBluePlus.startScan`) right after init.

The crash is reported on whichever method happens to be first to trigger `CBCentralManager` initialization — typically `getSystemDevices`. Stack trace shows the failure inside `-[FlutterBluePlusPlugin handleMethodCall:result:]`:

```
PlatformException(iosException, -[NSNull boolValue]: unrecognized selector sent to instance ...
  0 CoreFoundation
  1 libobjc.A.dylib         objc_exception_throw + 88
  2 CoreFoundation
  3 CoreFoundation
  4 CoreFoundation          _CF_forwarding_prep_0 + 96
  5 flutter_blue_plus_darwin -[FlutterBluePlusPlugin handleMethodCall:result:] + 672
  ...
)
```

## Root cause

Flutter engine 3.10+ (Dart 3.10+) changed how `MethodChannel`'s `StandardMessageCodec` encodes Dart `null` values inside `Map<String, dynamic>`:

* **Before:** Dart `null` → ObjC `nil` (key absent or value is nil)
* **Now:** Dart `null` → ObjC `[NSNull null]` (singleton, present in the dictionary)

The current `setOptions` handler in `FlutterBluePlusPlugin.m` uses `!= nil` to filter:

```objc
NSDictionary *args = (NSDictionary*) call.arguments;
if (args[@"show_power_alert"] != nil) {           // ← NSNull != nil is TRUE
    self.showPowerAlert = args[@"show_power_alert"];   // ← stores NSNull into NSNumber*
}
if (args[@"restore_state"] != nil) {
    self.restoreState = args[@"restore_state"];
}
```

`[NSNull null] != nil` is **TRUE**, so `NSNull` gets assigned to `self.showPowerAlert` (declared as `NSNumber*`). The crash happens later in the adapter init block when these properties are dereferenced:

```objc
// L141-150 in FlutterBluePlusPlugin.m
if ([self.showPowerAlert boolValue]) { ... }      // ← 💥 NSNull doesn't respond to boolValue
if ([self.restoreState boolValue]) { ... }
```

Because adapter init is lazy (only on first method call), the **crash surfaces on whichever method first triggers it**, not on `setOptions` itself. Common symptom: `getSystemDevices` / `systemDevices` `PlatformException`.

## Why Android is unaffected

Android's `StandardMessageCodec` on the Kotlin side returns Java `null` for absent / null values; Kotlin code at the plugin uses safe casts (`as? Boolean`), so a `null` is handled gracefully. iOS doesn't have an analogous safe-cast path — the value is just typed `id` and we lose the type guard.

## Reproduction

1. Use any Flutter app with `flutter_blue_plus: ^2.3.4` on Flutter 3.41+.
2. Call `await FlutterBluePlus.setOptions()` early (e.g. inside `main()` or page init). Note: `setOptions()` has all-optional `bool?` parameters; if user doesn't pass them, they default to `null`.
3. After that, call any BLE method that triggers adapter init (e.g. `systemDevices`, `adapterState`, `startScan`).
4. App crashes with the stack trace above.

Minimum repro on Dart side:
```dart
await FlutterBluePlus.setOptions(); // all defaults = nulls
final devs = await FlutterBluePlus.systemDevices(); // 💥
```

## Workaround (verified)

Patch `FlutterBluePlusPlugin.m` `setOptions` handler to reject `NSNull`:

```objc
if ([@"setOptions" isEqualToString:call.method]) {
    NSDictionary *args = (NSDictionary*) call.arguments;
    id powerAlert = args[@"show_power_alert"];
    if (powerAlert != nil && ![powerAlert isKindOfClass:[NSNull class]]) {
        self.showPowerAlert = powerAlert;
    }
    id restoreState = args[@"restore_state"];
    if (restoreState != nil && ![restoreState isKindOfClass:[NSNull class]]) {
        self.restoreState = restoreState;
    }
    result(@YES);
    return;
}
```

PR with the fix: incoming (link will be added here).

## Related

* Similar pattern was fixed in #1186 for `FlutterStandardTypedData`. This is the same class of bug — codec-shipped non-nil ObjC object being treated as a typed value.
* iOS-only issue #1163 has same surface symptom (`unrecognized selector`) but different root cause.

SmallBlueWhale:
Fixes #1333

## Problem

On Flutter 3.10+ / Dart 3.10+, the engine's `StandardMessageCodec` encodes Dart `null` values inside a `Map<String, dynamic>` as `[NSNull null]` instead of being absent from the resulting `NSDictionary`.

`setOptions`'s handler treats `args[@"show_power_alert"] != nil` as a truthy check, but `NSNull != nil` is TRUE — so `[NSNull null]` is stored into `self.showPowerAlert` (typed `NSNumber*`). The crash then materializes on the **first BLE method call** that triggers `CBCentralManager` init, when `[self.showPowerAlert boolValue]` is invoked on the `NSNull` singleton:

```
-[NSNull boolValue]: unrecognized selector sent to instance ...
  5 flutter_blue_plus_darwin -[FlutterBluePlusPlugin handleMethodCall:result:] + 672
```

User-facing symptom is usually `PlatformException` on `getSystemDevices` / `adapterState` / `startScan` — but the actual fault is in `setOptions` stashing `NSNull` into a property earlier.

Android is not affected because Kotlin's safe-cast `as? Boolean` returns `null` for non-Boolean values, including Java `null`.

## Fix

Reject `NSNull` explicitly with `isKindOfClass:[NSNull class]` before assigning. This keeps the property `nil` (its default), and `[nil boolValue]` is a valid ObjC no-op returning 0.

```diff
-    if (args[@"show_power_alert"] != nil) {
-        self.showPowerAlert = args[@"show_power_alert"];
-    }
-    if (args[@"restore_state"] != nil) {
-        self.restoreState = args[@"restore_state"];
-    }
+    id powerAlert = args[@"show_power_alert"];
+    if (powerAlert != nil && ![powerAlert isKindOfClass:[NSNull class]]) {
+        self.showPowerAlert = powerAlert;
+    }
+    id restoreState = args[@"restore_state"];
+    if (restoreState != nil && ![restoreState isKindOfClass:[NSNull class]]) {
+        self.restoreState = restoreState;
+    }
```

## Why minimal patch

* I considered a generic helper (e.g. `safeGet:` that filters `NSNull`), but `setOptions` is the only handler in this file that pulls "typed" values from `args` via `!= nil`. Other handlers cast via `requiredDouble` / `requiredString` etc. which already type-check.
* Same shape as #1186's `FlutterStandardTypedData` fix — codec-shipped non-nil sentinel handled at the use site.

## Verification

* Reverted patch → crash reproducible on Flutter 3.41.2 / iOS 17, 18 with:
  ```dart
  await FlutterBluePlus.setOptions(); // all nulls
  await FlutterBluePlus.systemDevices; // crashes
  ```
* Applied patch → same code path now succeeds; `setOptions(showPowerAlert: true)`, `setOptions(showPowerAlert: false)`, and `setOptions()` (all defaults) all behave correctly.
* No behavior change on Flutter <3.10 (codec there shipped Dart `null` as ObjC `nil`, which the old `!= nil` check already filtered correctly).

## Files changed

* `packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m`

### DIFF
--- a/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
+++ b/packages/flutter_blue_plus_darwin/darwin/flutter_blue_plus_darwin/Sources/flutter_blue_plus_darwin/FlutterBluePlusPlugin.m
@@ -112,11 +112,20 @@ typedef NS_ENUM(NSUInteger, LogLevel) {
         if ([@"setOptions" isEqualToString:call.method])
         {
             NSDictionary *args = (NSDictionary*) call.arguments;
-            if (args[@"show_power_alert"] != nil) {
-                self.showPowerAlert = args[@"show_power_alert"];
+            // Flutter 3.10+ MethodChannel codec encodes Dart `null` as
+            // `[NSNull null]`, not `nil`. The old `args[key] != nil` check is
+            // TRUE for NSNull, so NSNull gets stored into the NSNumber* properties
+            // and later `[self.showPowerAlert boolValue]` (during adapter init)
+            // crashes with `-[NSNull boolValue]: unrecognized selector`.
+            // Explicitly reject NSNull so the property stays nil and ObjC's
+            // nil-messaging gives the expected 0 return.
+            id powerAlert = args[@"show_power_alert"];
+            if (powerAlert != nil && ![powerAlert isKindOfClass:[NSNull class]]) {
+                self.showPowerAlert = powerAlert;
             }
-            if (args[@"restore_state"] != nil) {
-                self.restoreState = args[@"restore_state"];
+            id restoreState = args[@"restore_state"];
+            if (restoreState != nil && ![restoreState isKindOfClass:[NSNull class]]) {
+                self.restoreState = restoreState;
             }
             result(@YES);
             return;


chipweinberger:
no AI slop please

PR #1339, opened 2026-08-07
Darwin: only poll for MTU changes during connection

cspensky:
…change

The MTU poll was armed the first time the plugin touched the central manager and never invalidated, so an app that had used Bluetooth once woke up 40 times a second for the rest of the process lifetime, with nothing connected.

The ATT MTU is negotiated once per connection and is immutable afterwards, so there is only ever a short window worth observing. Arm the timer in didConnectPeripheral instead, and let checkForMtuChangesCallback disarm it once every connected peripheral's MTU has been unchanged for 2s. A new connection re-arms it.

Also drops the poll from 40Hz to 10Hz. The exchange completes well under a second, so the extra 30 wakeups a second bought nothing.

This was causing major resource drain on our phones in practice when the app was connected over BLE in the background.


chipweinberger:
```
        // Note: the MTU poll is armed on connect (see didConnectPeripheral), not here. Arming it
        // here left it running for the process lifetime even with nothing connected.
```

this comment no longer makes sense since the code it is referring to was deleted.

chipweinberger:
this is a good, reasonable improvement. thanks.

chipweinberger:
Released in 2.3.12

